Issue #4, opened 2018-01-26
In Modes 0 - 6, pick offset that is at the centre of the minima


hoglet67:
In Mode 0-6 here's an example calibration:
```
INFO: Calibrating modes 0..6
DEBUG: sample point 0: diff:  sum = 0 mean = 0, min = 0, max = 0
DEBUG: sample point 1: diff:  sum = 9280 mean = 928, min = 86, max = 1842
DEBUG: sample point 2: diff:  sum = 7132 mean = 713, min = 330, max = 2486
DEBUG: sample point 3: diff:  sum = 5668 mean = 566, min = 290, max = 884
DEBUG: sample point 4: diff:  sum = 0 mean = 0, min = 0, max = 0
DEBUG: sample point 5: diff:  sum = 0 mean = 0, min = 0, max = 0
INFO: Setting sp_default = 0
```
Now, you can see at several positions the sampling is error free. In this case an offset of 0 is picked for the default sampling point (sp_default), because none of the later positions improve on that.

There is actually room for improvement here. If you re-order the measurements, it's cleared that actually offset 5 would have been a better choice, because it's in the centre of the minima:
```
DEBUG: sample point 2: diff:  sum = 7132 mean = 713, min = 330, max = 2486
DEBUG: sample point 3: diff:  sum = 5668 mean = 566, min = 290, max = 884
DEBUG: sample point 4: diff:  sum = 0 mean = 0, min = 0, max = 0
DEBUG: sample point 5: diff:  sum = 0 mean = 0, min = 0, max = 0
DEBUG: sample point 0: diff:  sum = 0 mean = 0, min = 0, max = 0
DEBUG: sample point 1: diff:  sum = 9280 mean = 928, min = 86, max = 1842
```

hoglet67:
This was implemented by:
98f62373

Issue #1, opened 2018-01-25
Idea: use EDID for automatic adjustment to monitor geometry


BigEd:
If it can be done in bare metal, perhaps read the monitor's capabilities by EDID and adjust automatically?

In linux, you'd use `tvservice -d edid.dat ; edidparser edid.dat`

Perhaps TAG_GET_EDID_BLOCK in rpi-mailbox-interface.h is a clue.

The hope being that this removes the need to adjust the config file for different monitors.

hoglet67:
Yes, that would be a great feature.

What I'm not sure about is how easy it is to change the actual HDMI screen mode that's being used programatically. Currently this is set in config.txt, e.g.:
```
# 1920x1080p @ 50Hz
hdmi_group=1
hdmi_mode=31
overscan_top=-48
overscan_bottom=-48
overscan_left=240
overscan_right=240
framebuffer_aspect=-1
```
Dave

BigEd:
So do we need a baremetal equivalent to `tvservice --explicit=<string>`?
Ref:
https://raspberrypi.stackexchange.com/a/40681

hoglet67:
This looks relevant:
https://github.com/raspberrypi/firmware/issues/637

BigEd:
Good find!

PR #11, opened 2018-11-28
Update capture_line_default_4bpp.S

dp111:
simplified code slightly

### DIFF
--- a/src/capture_line_default_4bpp.S
+++ b/src/capture_line_default_4bpp.S
@@ -22,7 +22,9 @@ capture_line_default_4bpp:
 
         push    {lr}
         mov     r6, #0
-        ldr     r7, =0x11111111
+        mov     r7, #0
+        tst     r3, #BIT_VSYNC_MARKER
+        ldrne   r7, =0x11111111
 loop:
         WAIT_FOR_PSYNC_10                // expects GPLEV0 in r4, result in r8
 
@@ -33,9 +35,8 @@ loop:
         CAPTURE_HIGH_BITS                // input in r8, result in r10, corrupts r9/r14
 
         // Orr in the VSync indicator
-
-        tst    r3, #BIT_VSYNC_MARKER
-        orrne  r10, r10, r7
+        
+        orr  r10, r10, r7
 
         // Line double always in Modes 0-6 regardless of interlace
         // On the multi core Pi this introduces stalling artefacts
--- a/src/capture_line_default_4bpp.S
+++ b/src/capture_line_default_4bpp.S
@@ -21,12 +21,9 @@
 capture_line_default_4bpp:
 
         push    {lr}
-
-        mov    r12, r0                   // pointer to the line in the frame buffer
-        mov    r6, r1                    // number of 8-bit characters per line
-
+        mov     r6, #0
+        ldr     r7, =0x11111111
 loop:
-
         WAIT_FOR_PSYNC_10                // expects GPLEV0 in r4, result in r8
 
         CAPTURE_LOW_BITS                 // input in r8, result in r10, corrupts r9/r14
@@ -36,20 +33,19 @@ loop:
         CAPTURE_HIGH_BITS                // input in r8, result in r10, corrupts r9/r14
 
         // Orr in the VSync indicator
-        ldr    r8, =0x11111111
+
         tst    r3, #BIT_VSYNC_MARKER
-        orrne  r10, r10, r8
+        orrne  r10, r10, r7
 
         // Line double always in Modes 0-6 regardless of interlace
         // On the multi core Pi this introduces stalling artefacts
 #ifndef HAS_MULTICORE
         tst    r3, #BIT_SCANLINES
-        movne  r0, #0
-        moveq  r0, r10
-        str    r0, [r12, r2]
+        streq  r10, [r0, r2]
+        strne  r6, [r0, r2]
 #endif
-        str    r10, [r12], #4
-        subs   r6, r6, #1
+        str    r10, [r0], #4
+        subs   r1, r1, #1
         bne    loop
 
         pop    {pc}
--- a/src/capture_line_default_8bpp.S
+++ b/src/capture_line_default_8bpp.S
@@ -74,8 +74,8 @@ loop:
         streq  r10, [r0, r2]
         strne  r6, [r0, r2]
 #endif
-        str    r10, [r0], #4
         subs   r1, r1, #1
+        str    r10, [r0], #4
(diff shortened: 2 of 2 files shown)


hoglet67:
Hi Dom,

Thanks for this. I've just tested it and it seems there is a problem with the change to the 8bpp file. Specifically, it's a performance problem rather than a coding problem.

I've tried to narrow it dowm, and it seems that changing from:

```
        tst    r3, #BIT_SCANLINES
        movne  r6, #0
        moveq  r6, r10
        str    r6, [r0, r2]
        str    r10, [r0], #4
```
to
```
        tst    r3, #BIT_SCANLINES
        strne  r6, [r0, r2]
        streq  r10, [r0, r2]
        str    r10, [r0], #4
```
seems to cause more stalling such that sampling jitter becomes apparent.

I did wonder if the the existing code was marginal, so I changed:
```
        push    {lr}
```
to
```
        push    {r0-r12, lr}
```
If it were marginal, that extra stacking of registers on entry might be expected to push it over the edge. It didn't though.

Any thoughts?

hoglet67:
Here's another data point.

This suffers from slightly worse sampling jitter:
```
capture_line_default_8bpp:

        push    {lr}

        mov    r12, r0                   // pointer to the line in the frame buffer
        mov    r6, r1                    // number of 8-bit characters per line
loop:

        WAIT_FOR_PSYNC_10

        CAPTURE_BITS

        // Line double always in Modes 0-6 regardless of interlace
        // On the multi core Pi this introduces stalling artefacts
        str    r10, [r12, r2]
        str    r10, [r12], #4

        WAIT_FOR_PSYNC_01

        CAPTURE_BITS

        // Line double always in Modes 0-6 regardless of interlace
        // On the multi core Pi this introduces stalling artefacts
        str    r10, [r12, r2]
        str    r10, [r12], #4

        subs   r6, r6, #1
        bne    loop

exit:
        pop    {pc}
```
Where as my original version, with additional instructions intermingled, doesn't.

hoglet67:
And another data point - it only happens with H_OFFSET being an even value. This is looking like a bug where I'm starting on the wrong edge of PSYNC. I'll check the code....

dp111:
I would be interested to see if this is any better.

capture_line_default_4bpp:
push {lr}
mov r6, #0
mov r7, #0
tst r3, #BIT_VSYNC_MARKER
ldrne r7, =0x11111111
loop:
WAIT_FOR_PSYNC_10 // expects GPLEV0 in r4, result in r8
CAPTURE_LOW_BITS // input in r8, result in r10, corrupts r9/r14

orr r10, r10, r7
WAIT_FOR_PSYNC_01 // expects GPLEV0 in r4, result in r8
CAPTURE_HIGH_BITS // input in r8, result in r10, corrupts r9/r14
// Line double always in Modes 0-6 regardless of interlace
// On the multi core Pi this introduces stalling artefacts
#ifndef HAS_MULTICORE
tst r3, #BIT_SCANLINES
streq r10, [r0, r2]
strne r6, [r0, r2]
#endif subs r1, r1, #1
str r10, [r0], #4
bne loop
pop {pc}

hoglet67:
Dom,

I'm slightly confused - the performance problem was only with the 8bpp version. You've just switched to the 4bpp version.

Dave

dp111:
sorry

See my branch patch1 for a possible new version to try.

hoglet67:
Can't see anything on patch-1:
https://github.com/dp111/RGBtoHDMI/commits/patch-1

Did you push?

dp111:
Sorry again patch2 branch

hoglet67:
Your last change in patch-2 made no difference to the sampling jitter.

However,  going back to the previous version in patch-2 and adding two consecutive NOPs pretty much anywhere in the loop removes the sampling jitter (one NOP is not sufficient).

That doesn't make a lot of sense to me, because the loop is paced anyway by psync edges.

I wonder if the store buffer is very close to overflowing, and adding the NOPs just prevents that happening, thus avoiding a much more expensive pipeline stall.

hoglet67:
This also seems to work:
```
// The capture line function is provided the following:
//   r0 = pointer to current line in frame buffer
//   r1 = number of 8-pixel blocks to capture (=param_chars_per_line)
//   r2 = frame buffer line pitch in bytes (=param_fb_pitch)
//   r3 = flags register
//   r4 = GPLEV0 constant
//   r5 = frame buffer height (=param_fb_height)
//   r6 = scan line count modulo 10
//
// All registers are available as scratch registers (i.e. nothing needs to be preserved)

capture_line_default_8bpp:

        push    {lr}

        mov     r7, #0
        tst     r3, #BIT_VSYNC_MARKER
        ldrne   r7, =0x01010101

        add     r2, r0
loop:   
        WAIT_FOR_PSYNC_10

        CAPTURE_BITS

        // Orr in the VSync indicator
        orr  r5, r10, r7

        WAIT_FOR_PSYNC_01

        CAPTURE_BITS
   
        // Orr in the VSync indicator
        orr    r10, r10, r7

        str    r5,  [r0], #4
        str    r10, [r0], #4
   
        tst    r3,  #BIT_SCANLINES  
        movne  r5,  #0
        movne  r10, #0
        str    r5,  [r2], #4
        str    r10, [r2], #4

        subs   r1, r1, #1
        bne    loop

exit:
        pop    {pc}
```

dp111:
Its complex.

in the last case the eq case has two cycles between the executed stores.
the ne case I suspect gets combined in the write buffer into one. ne eq ne
eq is too spaced out to combine ?

Try using STM

        stmia    r0!,{r5, r10}

        tst    r3,  #BIT_SCANLINES
        movne  r5,  #0
        movne  r10, #0

        stmia   r2!,{r5, r10}

hoglet67:
Using STM works as well.

I also added 100 NOPs after the stores and that didn't cause it to break. So I don't think the loop as a whole is marginal.

Anyway, this is useful for understanding (or not!).

But the problem I would really like to solve is more efficient pattern for rendering the OSD.

In modes 0..6 we have to resort to redraw the OSD each frame, in the vertical blanking. Which means we must use multiple buffering, or it flickers really badly.

What I was thinking of trying was rending the OSD once to normal cacheable memory, and then trying to merge this into the line as the capture is happening. Odd and even lines have different OSD data, so there might need to be two read pointers, but they will both be reading sequentially. Do you think there is any chance that would work?

dp111:
Given there is 100 spare cycles , I think putting the OSD in cacheable and
merging should be easily possible.  Also I think we have r11 and r12 free
so we aren't short of registers

PR #15, opened 2018-11-30
Update capture_line_default_8bpp.S

dp111:
Free up R14 in CAPTURE_BITS


hoglet67:
Dom,

Thanks - I'd already spotted and made this change.

FYI, I've just done another push that has reworked the WAIT_FOR_PSYNC_ macros. It adds an three extra EOR instructions, but saves a couple of branches.

More importantly, it frees the capture line code from having to track the polarity of the next edge.

Dave

PR #18, opened 2018-12-02
Update capture_line_mode7_4bpp.S

dp111:
Untested Optimisations


hoglet67:
FYI, there were some conflicts between patch-8 and patch7, because patch-8 was based on patch-6.

I think I resolved them, and the result seems to work.

Dave

PR #19, opened 2018-12-03
Genlock changes and bug fixes

IanSB:
I've made some changes and bug fixes to the genlock introducing some hysteresis and also restricted the range to 5 - 265 to allow that to work properly. (The single buffer OSD only starts to disappear around 20-25 so that shouldn't be an issue)
I also changed cmdline.txt to accomodate the above and add bob deinterlace.
Note I have changed the hdmi clock logging to log_debug as this was causing frame drops whenever there was a resync although I have also added some extra logging to see how frequent the resyncs are happening which may still cause frame drops.
It may be better in the long term to pre-calculate the hdmi pll values and maybe re-implement 'recalculate_hdmi_clock_line_locked_update()' in assembler using those values to eliminate any chance of field drops during a resync.

### DIFF
--- a/src/osd.c
+++ b/src/osd.c
@@ -114,9 +114,9 @@ static param_t features[] = {
    {   F_SCANLINES,       "Scanlines", 0,                    1, 1 },
    {         F_ELK,             "Elk", 0,                    1, 1 },
    {         F_MUX,       "Input Mux", 0,                    1, 1 },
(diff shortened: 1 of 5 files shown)


IanSB:
Reduce hysteresis range. To increase value we will need to reduce genlock range from 5 - 265

PR #29, opened 2019-02-05
16 colour mode 2

IanSB:
This is my first attempt at 16 colour mode 2 with in band palette update. Demo disk here:
[hdmidemo.zip](https://github.com/hoglet67/RGBtoHDMI/files/2831359/hdmidemo.zip)

Notes:
You have to get mode 0 to align with the start of the 3rd byte (switch on debug for alignment markers)
The alignment setting works in mode 0 but interacts with the sample setting so there may be an issue with the CPLD. (My delay setting was 6)
Detection is always on at the moment so might be triggered by random screen data.
More work needs to be done on defining an in band protocol that can work across all modes.
Also there is an issue where if reset is pressed with in band data on screen, the Pi gets messed up about 50% of the time (maybe trying to update the palette before other things have been initialised?)

### DIFF
--- a/src/capture_line_default_4bpp.S
+++ b/src/capture_line_default_4bpp.S
@@ -6,6 +6,7 @@
 .text
 
 .global capture_line_default_4bpp
+.global inBandData
 
 // The capture line function is provided the following:
 //   r0 = pointer to current line in frame buffer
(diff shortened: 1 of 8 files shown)


IanSB:
Increased buffer sizes as reset problem above might be buffer overrun.
Also is CPLD delay value working with CLK/8 instead of CLK/6 in modes 0-6? (Would explain the interaction with sample setting)

PR #30, opened 2019-03-04
8 bit capture, 16x16 palettes & half resolution support

IanSB:
You might want to start reviewing and merging these changes although they are not complete as I still have to implement the full command side of the in band control.
The capture side has been completely rewritten, the in band data will work anywhere on the screen and supports multiple in band lines. 8 bit capture is now the default for modes 0-6 so the menus work OK with 16 colours and back end support for switching palettes on a per line basis has been added (control commands not yet implemented).  The existing 16 colour demo should still work.
Also half resolution support has been added using the Half Odd and Half Even sampling options. The menu system has been tweaked and most menus are OK in half resolution but some more work needs to be done there to clean up remaining issues.  
I used the following command line for testing half resolution:
sampling06=2,2,2,2,2,2,2,0,6,5 sampling7=0,3,3,3,3,3,3,0,8,4 geometry06=32,21,84,270,336,270,8,96000000,6144,5000,3 info=1 palette=0 deinterlace=6 vsync=0 vlockmode=3 nbuffers=0


hoglet67:
Hi Ian,

I'll give this some attention this week.

One initial observation...

The default capture code is now much more complicated, in that it is always running some of the in-band sentinel detection code (though I appreciate if BIT_IN_BAND_ENABLE is 0 then the result will always be false).

Did you consider leaving the original minimal capture code as-is, and adding a new version that includes the in-band sentinel detection along side as a separate file (and enabled with a config  option)?

I have several concerns with not doing this:

1. The more complex capture code is being used all the time, even for use cases that have no need of it.

2. If people have any issues in the future, it's not now possible to revert to to the simple/well proven capture code.

3. There may be some cases where the previous capture code would have worked (e.g. for a higher pixel clock) that the new code doesn't cope with.

4. Any capture issues (e.g. caching related) that arise in the future will be harder to debug, because we can't fall back to the old version. With PiTubeDirect, before the ULA emulation was run in the GPU, we had some very nasty caching issue. We thought we had the system stable, and then adding unrelated code would mess things up, because everything gets shifted around by the compiler.

Generally, the thing I was trying to achieve with by factoring out the capture code as I had was to make the whole system a bit more modular and plug-able.  i.e. If some custom/experimental capture code was developed, then this could be run along side the existing capture code, and enabled by an option in the config file.

Dave

IanSB:
There are now the following capture files:
capture_line_inband_4bpp_8bpp.S (Used for normal, odd & even with in band data)
capture_line_default_4bpp_8bpp.S (optimised 4 and 8 bit capture)
capture_line_oddeven_4bpp_8bpp.S
capture_line_half_4bpp_8bpp.S
The last 2 are not fully optimised as they can do both odd and even processing so they could be split further at a later date but it would only save a few register based instructions.

hoglet67:
Thanks for doing that change Ian.

I'll have a good look/play with this tomorrow.

IanSB:
Noticed a few bugs with geometry handling:
1. adjusting the FB_height in the menu trashes the screen
2. setting V height or H width larger than the defaults in a command line doesn't work because those values are limited by the old FB width/height in update_param_range() before the new FB height/width are read from the command line.  Maybe need to read those values in reverse order to fix it.
3. Auto calibration messes the screen up at half resolution with 8 bits/pixel using the following half res BBC geometry (It seems to work OK at 4 bits/pixel):
geometry06=32,21,42,270,336,270,8,96000000,6144,5000,6

IanSB:
Could do with another option in the sampling menu / command line:
V Delay: This is the offset from the top of the frame buffer to the first active line of video captured. This is used by mode 7 deinterlacing to determine the right line pair and now it's used in modes 0-6 by the multi palette to determine which line to apply the selected palette. It's currently hard coded in both cases but would be wrong if some of the vertical parameters were changed.

hoglet67:
Ian,

I'm just jotting down some notes here as I work through testing and reviewing the changes.

** Testing of a normal sized frame buffer**

The Auto Calibration messages (and other notifications) have moved down the screen by two lines.

Changing the FB Height causes things to mess up.

Auto Calibration in even/odd mode seems to revert to normal mode during calibration. That may be by design?

Setting half-even or half-odd sampling mode causes screen to flash. See below comments on the purpose of these sampling modes?

** Testing of a low-res frame buffer

(I used  your cmdline.txt example above)

The top pixel row of the OSD (e.g. the bar at the top of the "I" in Info Menu") is being blanked. This only happens in single buffered mode.

Changing the FB Height or FB Width causes a crash.

The calibration code no longer manages to avoid the lines containing the flashing cursor, so generally you see some errors.

Auto Calibration in half mode seems to mess up (debug gets enables during calibration).

**Review: Capture Code:**

- We should get rid of HAS_MULTICORE - the multi core Pi will never was, and probably never will be, supported as it's memory system has much longer/more variable latency. This macro just makes reading the code harder.

Looking at the sampling mode changes I have a couple of questions:
``
+   PS_INBAND,    // Each sampled pixel is mapped to one pixel in the frame buffer + In band data
+   PS_INBAND_E,  // Even pixels are replicated, odd pixels are ignored + In band data
+   PS_INBAND_O,  // Odd pixels are replicated, even pixels are ignored + In band data
    PS_NORMAL,    // Each sampled pixel is mapped to one pixel in the frame buffer
-   PS_SUBSAMP_E, // Even pixels are replicated, odd pixels are ignored
-   PS_SUBSAMP_O, // Odd pixels are replicated, even pixels are ignored
-   PS_DOUBLE,    // Each sampled pixel is mapped to two pixels in the frame buffer
+   PS_NORMAL_E,  // Even pixels are replicated, odd pixels are ignored
+   PS_NORMAL_O,  // Odd pixels are replicated, even pixels are ignored
+   PS_HALF_E,    // Even pixels are used, odd pixels are ignored
+   PS_HALF_O,    // Odd pixels are used, even pixels are ignored
``
- What's happened to the pixel doubling sampling mode (capture_line_default_4bpp_double.S). This was used on the Atom to allow it to run at 57.27MHz clock with a full (i.e. 2x) resolution frame buffer.

- Are the PS_HALF_E and  PS_HALF_O actually necessary. What's the advantage over running with PS_NORMAL and halving the speed of the pixel clock?

i.e. geometry06=5,21,42,270,336,270,8,48000000,3072,5000,3

[ Digression: It would be good to write a end-user understandable description of the different sampling modes. I'm also wondering if it's right to enable/disable the inband commands via the samping mode option, or whether then should be an orthogonal flag. I guess it depends if it really is orthogonal? (i.e. it would have to be available in all sampling modes. ]

- I like making the line doubling programmable, via BITDUP_NO_LINE_DOUBLE, but is seems wrong to be enabling this based on a comparison with a hard coded constant DUPLICATE_HEIGHT (380). Why 380 specifically? Can't this just be made a geometry option?

[ Digression: I'm not particularly happy with the use of cmdline.txt for params, and especially the use of comma separated lists. Dominic has used this library (together with fatfs) in the Pi 1MHz bus work to allow access to the SD Card:
https://github.com/jncronin/rpi-boot/blob/master/emmc.c
I'm going to look at integrating this at some point, then we can do things like multiple custom config files, saving of config files, screen capture, and possibly even recording of short video segments (buffered via RAM).

- Let's change the "HDMI" sentinel to something else. I am slightly paranoid about HDMI licensing.

- Do the comments documenting the interface to the capture line functions ("The capture line function is provided the following") need updating to include additional registers? ('m not clear if the additions to rgb_to_fb.S are effectively adding more params or not.

**Review: C Code:**

- In geometry.c the default for mode 7 is:
mode7_geometry.px_sampling   = PS_INBAND;
Does this make sense? I didn't think in band data could be supported in mode 7.

**Review: Other**

- The documentation in cmdline.txt for sampling modes and needs updating.

- Also, references to 1000ppm should now be 500ppm.

Overall, I wonder if it's worth using a smaller OSD font for the reduced-size frame buffer modes?

Thanks for all your effort on this Ian, it's much appreciated and you are doing some great work here.

IanSB:
> The Auto Calibration messages (and other notifications) have moved down the screen by two lines.
That was so they are visible in half height modes (top two lines are dropped to make menus visible)
> What's happened to the pixel doubling sampling mode 
Forgot to add that back in.
> What's the advantage over running with PS_NORMAL and halving the speed of the pixel clock
Didn't realise the clock could be set that low. I'll try it with the ZX81 and if it looks OK I'll remove the half settings
> hard coded constant DUPLICATE_HEIGHT (380). Why 380 specifically?
EGA is 350 pixels high and doesn't need duplication and the lowest vertical resolution of 15Khz modes I have seen is 192 pixels which doubles to 384 so 350 is roughtly half way between 350+padding and 192*2+padding and I was trying to avoid adding yet another command line parameter / menu option.
> mode7_geometry.px_sampling = PS_INBAND;
I think that line is ignored in mode 7 but it was changed to PS_INBAND as that was the zero enum however as it is visible in the menu it should be changed back.
>  Overall, I wonder if it's worth using a smaller OSD font for the reduced-size frame buffer modes?
I think it is now going to be necessary as there are going to be quite a few additional menu options which will overflow the screen in half height mode so I'll revert all the cosmetic OSD changes. Is there an 8x8 font available in the same format as the mode 7 font?

As overscan can be programmed via the mailbox, I think that some scaling options also need to be added to the menus as that is one area that is very non-intuitive due to the interaction between the screen size, the overscan values and the frame buffer size. Perhaps something like:
Scaling:  Integer / Non Integer
H Border: 0-n
V Border: 0-n
If Scaling is set to Integer, the H and V border settings should jump to discrete values that result in integer scaling based on current screen and buffer size.
(Separate settings would be needed for mode7 / modes0-6)

IanSB:
> Are the PS_HALF_E and PS_HALF_O actually necessary. What's the advantage over running with PS_NORMAL and halving the speed of the pixel clock? i.e. geometry06=5,21,42,270,336,270,8,48000000,3072,5000,3

I tried this and it just about works but the calculated nominal core clock is 400Mhz which is the upper limit so there is no headroom for locking the the source if the core needs to be increased.
Also I tried an equivalent ZX81 setting but that produced a very messed up screen result although the calculated clock was 390Mhz so maybe that issue lies elsewhere
Works:
geometry=11,21,40,270,320,270,4,78000000,4968,0,6
Doesn't work:
geometry=11,21,40,270,320,270,4,39000000,2484,0,3

Also the menu system in the current code limits the clock to 75Mhz

hoglet67:
The limits are a bit arbitrary and the values predate much of our recent work.

The code in rgb_to_hdmi calibrate_sampling_clock() possibly needs tweaking to pick the right divisor:

For a 48MHz Clock.  a divisor of 25 implies a core clock of: 48 * 25 / 3 = 400MHz. Which, as you say, hits our arbitrary 400MHz upper limit. A divisor of 24 would give work better and give 384MHz.

For a 39MHz Clock, a divisor of 30 implies a core clock of: 39 * 30 / 3 = 390MHz. This should have worked.

I'll investigate this second case a bit more, see what's actually being fed into the CPLD.

Dave

hoglet67:
I've had a think about the 39MHz case.

I've verified the clock really is 39Mhz, so it's not a PLL issue.

I think what's messing up is the way the CPLD decides where to start sampling the line, based on a fixed number of clocks from the falling edge of HSYNC. The lower the clock, the longer this delay is going to be. And depending on H Width, there is then a danger of the software missing the next HSYNC pulse.

The value in the CPLD is currently 832.

- At 96MHz and a CPLD divisor of 6 this corresponds to 832 * (6/8) / 96 = 6.5us. So that starts sampling ~1.5us after the 5us sync pulse.
- At 78MHz. this increases to 832 * (6/8) / 78 = 8.0us, so that's ~3us after the sync pulse ends.
- At 39MHz. this increases to 832 * (6/8) / 39 = 16.0us, so that's ~11us after the sync pulse ends.

So there is a possibility at 39MHz of missing the start of the signal.

Now lets look at the H offset + H width values.

- With a clock of 39MHz and a divisor of 6 there will be a psync clock every 6*4/39 = 0.615us.
- A H_Offset of 11 (units of psync clock edges) would correspond to 11 * 0.615 = 6.8us
- A H_Width of 40 (units of psync clock cycles) would correspond to 40 * 2 * 0.615 = 49.2us.
- 11 + 6.8 + 49.2 = 69us, which is longer than a line.

If you reduce H_Offset to 0 and H_Width to 38 then picture looks better, and the system is no longer missing sync pulses.

But the killer here is still the way the CPLD calculates sampling initlal sampling delay:
- a fixed number of sampling clock cycles
- from the start (i.e. the falling edge) of psync

I might do some experiments with using the other edge of hsync. I know this will break the smooth horizontal scrolling, but that might be fixable in software.

I would be interested in what the ZX81 looks like with an H_Offset of 0 and a H_Width of 38 (ish).

But I'd say for now keep your half rate sampling code.

IanSB:
I've added support in the CPLD for CGA style separate H & V sync. It works ok and doesn't affect normal BBC operation.
Are you working on any code at the moment or do you want me to make the changes discussed above before merging? If so do you know of an 8 bit font we could use for the menu?

hoglet67:
The CPLD change is fine, and is what I was thinking as well. Sorry I didn't get around to doing it. I've been working on a 1MHz Bus FPGA Adapter (which got sent to JLCPCB yesterday).

On the intensity bit change: what system(s) was this aimed at? What format did you imagine passing the data between the CPLD and the host (the bus is 12 bits wide). Three pixels at a time, rather than four?

> Are you working on any code at the moment or do you want me to make the changes discussed above before merging?

I'm not working on this at the moment, so please do keep on making changes. 

Once the dust settles, I was going to look at adding the EMMC library and doing a re-work of cmdline.txt into multiple config files. I have several goals here:
- making it easier for users to work with
- supporting different modes (or even machines) in different files
- manually toggling between the files
- persistently saving the default
- persistently saving the calibration data

I'm not going to start this until you are done with your current work,

> If so do you know of an 8 bit font we could use for the menu?

What about using the BBC font? This is at the start of the MOS ROM:
http://mdfs.net/Docs/Comp/BBC/OS1-20/C000

Dave

hoglet67:
With the OSD font, I was thinking we would keep the current 20x12 font for large frame buffers, and only fall back to the 8x8 if it's likely the OSD would otherwise overflow.

I'd also keep the 8x8 and 20x12 rendering code separate. The performance of the current 20x12 is actually quite close to the point at which the next VSYNC pulse might be missed (in Modes 0..6).

Re: merging, I'm also happy to merge what you have done so far back into dev, if you would like. Just let me know.

IanSB:
> With the OSD font, I was thinking we would keep the current 20x12 font for large frame buffers, and only fall back to the 8x8 if it's likely the OSD would otherwise overflow.
> I'd also keep the 8x8 and 20x12 rendering code separate. The performance of the current 20x12 is actually quite close to the point at which the next VSYNC pulse might be missed (in Modes 0..6).

That's what I intend to do. BTW the next VSYNC pulse is already missed with 8bits/pixel so I may rewrite the fast osd code in assembler in the future.

> Re: merging, I'm also happy to merge what you have done so far back into dev, if you would like. Just let me know.

Probably best to wait until the next update as I'll be reverting some changes.

> On the intensity bit change: what system(s) was this aimed at? What format did you imagine passing 
the data between the CPLD and the host (the bus is 12 bits wide). Three pixels at a time, rather than four?

IBM PC CGA (14.318 Mhz clock at 15Khz/60Hz RGBI)
There is also IBM PC EGA which is has a 16.257Mhz clock and 6 bits of RGB (2 each)
I was looking at the CPLD to see what was possible and it might be feasible to read 6 bits (3x RGB + 3xRGBMUX) by doubling the PSYNC rate and have 2 pixels per PSYNC edge. The amount of memory written would be the same as now for 8 bits/pixel but there would be double the PSYNC edges to read from the GPIO before writing a word. 
As a test I doubled the clock rate to 192Mhz and wrote another pixel loop to read 4 psyncs before writing a word and that amost worked. There was jitter but I'm not sure if it was CPU timing or running the CPLD too fast (I think it's rated for 100Mhz?). Rather than running the CPLD too fast it could be run at 96Mhz as now with double the PSYNC clock. This would mean only 3 sample positions instead of 6 but that may be enough for a non-mode 7 system.
Any thoughts?
A possible workaround for the memory write taking too long would be to use an asymmetric PSYNC clock with three fast edges and longer time after the fourth edge to allow the memory write to complete.

hoglet67:
> As a test I doubled the clock rate to 192Mhz and wrote another pixel loop to read 4 psyncs before writing a word and that amost worked. There was jitter but I'm not sure if it was CPU timing or running the CPLD too fast (I think it's rated for 100Mhz?). Rather than running the CPLD too fast it could be run at 96Mhz as now with double the PSYNC clock. This would mean only 3 sample positions instead of 6 but that may be enough for a non-mode 7 system.
> Any thoughts?

The current CPLD is almost certainly going to malfunction at 192MHz. It doesn't quite meet timing at 96MHz (at least, last time I checked). I can re-check this tomorrow.

For the  IBM PC EGA  16.257Mhz pixel clock, you could run the CPLD at 6x (97.542MHz) which gives you the 6 sample positions per pixel. PSYNC would need to toggle every two pixels (i.e. every 12 cycles). This is a slight variation over the current design, in that the PSYNC rate is ~twice as fast. But it should be feasible. The Atom CPLD worked a bit like this for a while (two 6-bit pixels per PSYNC)

> A possible workaround for the memory write taking too long would be to use an asymmetric PSYNC clock with three fast edges and longer time after the fourth edge to allow the memory write to complete.

I can see how this would help the software. But I think it implies far more registers than the CPLD has. i.e. you really need a small FIFO to buffer up several (4? 8?) pixels.

Dave

IanSB:
The new OSD auto switches between the old and new fonts. Note it currently uses 8x8 in full res 8 bits/pixel mode as the teletext font misses Vsync but that results in small menus. I tried doubling the height but that also missed Vsync.
I propose to add the following menu options unless you have any objections:
Geometry Menu:
FB height x 2: on/off
I think FB height should be specified as 270 rather than 540 and setting the above option to "on" doubles the actual size and enables the double writing bit so you only need to change one setting to switch between the two modes.
FB V offset: number of lines between top of frame buffer and start of actual video.
FB Scaling: Integer / Interpolated
Aspect ratio lock: off / 4:3 / 16:9 (maybe others)
V Border size: 0-n
H Border size: 0-n

Processing Menu:
Palette Control: Off / In Band Programming / NTSC Artifacting

Settings menu:
Computer type: Standard / BBC Micro / Electron
I propose to replace the Elk and Mode7 Disable switches with the above and that could be expanded as require for any other machines.

> PSYNC would need to toggle every two pixels (i.e. every 12 cycles). This is a slight variation over the current design, in that the PSYNC rate is ~twice as fast. But it should be feasible. 

It would be useful to get a CPLD with double the PSYNC rate to see if it's feasible to capture at full resoloution.

> I can see how this would help the software. But I think it implies far more registers than the CPLD has. i.e. you really need a small FIFO to buffer up several (4? 8?) pixels.

If it was a different build you could free up a lot of registers by using only 1 set of sampling points.

hoglet67:
> I think FB height should be specified as 270 rather than 540 and setting the above option to "on" doubles the actual size and enables the double writing bit so you only need to change one setting to switch between the two modes.

Agreed.

> FB Scaling: Integer / Interpolated

Is this allowing control of GPU scaling_kernel? How?

> Aspect ratio lock: off / 4:3 / 16:9 (maybe others)

Again, how is this implemented?

> FB V offset: number of lines between top of frame buffer and start of actual video.
> V Border size: 0-n
> H Border size: 0-n

This all seems to be aimed at allowing the active area of video to be captured to a window within the frame buffer. Is that correct?

How does FB V Offset differ from V Border size?

Is Border somehow symetrical?

At some point we are going to have to document all this. Maybe it would be good to try to write that descriptive text now. That's often a good way to test whether something really makes sense.

Will these be appended to the end of the current geometry list, with sensible defaults? This is where I think a new config file format is desperately needed. Maybe I'll take a look at the EMMC library today.

> Processing Menu:
> Palette Control: Off / In Band Programming / NTSC Artifacting
Can you say a bit more about NTSC Artifacting?

> 
> Settings menu:
> Computer type: Standard / BBC Micro / Electron
> I propose to replace the Elk and Mode7 Disable switches with the above and that could be expanded as require for any other machines.

I see the logic in this, but ultimately it's less flexible. How, for example, would it cope with an Electron with a Jafa mode 7 board?

> 
> > PSYNC would need to toggle every two pixels (i.e. every 12 cycles). This is a slight variation over the current design, in that the PSYNC rate is ~twice as fast. But it should be feasible.
> 
> It would be useful to get a CPLD with double the PSYNC rate to see if it's feasible to capture at full resolution.

Let me have a play over the next couple of days.

> 
> > I can see how this would help the software. But I think it implies far more registers than the CPLD has. i.e. you really need a small FIFO to buffer up several (4? 8?) pixels.
> 
> If it was a different build you could free up a lot of registers by using only 1 set of sampling points.

Again, lets me try and see what seems to fit.

Dave

IanSB:
> How does FB V Offset differ from V Border size?

FB Voffset is not related to the other proposed settings. The frame buffer is 270/540 high, the bbc mode screen is 256 high (500 high in mode 7) so a number of blank lines are captured at the top and bottom of the frame buffer. This value is the number of lines from the top of the frame buffer to the start of the actual bbc video (i.e. the number of blank lines at the top). This info is used by mode 7 deinterlace and the mode 0-6 16x16 palette and is currently hard coded. 

> Is this allowing control of GPU scaling_kernel? How?

No, the FB Scaling/ Aspect / Border were to automate the overscan settings. Perhaps FB Scaling should be Integer / Non Integer to avoid confusion with scaling kernel.
An alternative would be to use  H and V scale rather than H & V border but those values would only be whole numbers for integer scaling. The aspect ratio settings would only allow values that produced correct aspect ratio on the output screen.

> Can you say a bit more about NTSC Artifacting?

Composite NTSC computers can generate "artifact colours" by putting different high res patterns of pixels on the screen which appear as lower res coloured areas on NTSC monitors. This can be simulated by decoding the high res pixel patterns as larger coloured pixels which is very similar to decoding mode 0 as 16 colour mode 2. It's used by Apple II, Tandy color computer and IBM PC CGA. It's just a placeholder at the moment and can be commented out until implemented. More info here:
https://www.youtube.com/watch?v=niKblgZupOc
https://www.youtube.com/watch?v=yHXx3orN35Y

> I see the logic in this, but ultimately it's less flexible. How, for example, would it cope with an Electron with a Jafa mode 7 board?

Well it would just work as it does now. I was thinking of keeping the Elk and mode7disable flags internally and just presenting them in this manner as other settings might be needed e.g. for the ZX81.
Standard means Elk=0 and m7disable = 1
BBC means Elk=0 and m7disable = 0
Electron means Elk=1 and m7disable =0 (so jaffa board would work)
(You could add and "Alternative sync" option to cover all bases which would be Elk=1 m7disable=1

Let me know if you have any better suggestions for the menu names

IanSB:
Just had another thought on the FB V offset:
Instead it could be FB Active Video Height: 256 (250 for mode7 which would be doubled by the FB height x 2 setting). The software would then auto adjust things so that the video would be vertically centred in the buffer and it would then automatically know the offset value. This means that the vertical size of the buffer and  the Voffset could be adjusted without having any effect on deinterlace or palette 16x16 so you avoid having too many interacting adjustments. (With FB Voffset you would have to keep adjusting that to compensate for adjustments made to FB height)

hoglet67:
OK, consider just the vertical parameters for a moment, we have:
- V Offset (which controls the start of the capture within the field, i.e how many lines are skipped)
- V Height (which controls the number of lines captured)
- FB Height (which is the frame buffer height)
- FB Height x2 (on/off, which doubles the frame buffer height)
- FB V Offset (which is where to place the captured video within the frame buffer)
- V Border (which relates to automatically setting the overscan values, but I'm not understanding yet what it's doing)
- FB Scaling
- Aspect Ratio Lock
- plus the HDMI parameters set in config.txt which are currently fixed (e.g. the HDMI vertical resolution)

It does feel like we are possibly over complicating things here, and possibly need to step back and be clearer about the problem(s) we are trying to address with each of these.

I'm wondering if the FB Height could actually be set from the other parameters?

Dave

hoglet67:
If I'm understanding right, the only purpose of FB V Offset is to allow the software to extract information out of the video stream for:
- deinterlacing
- palette commands
- cursor rows (to avoid counting during calibration)

It seems that:
1. These are very Acorn specific things (or do you plan the palette stuff to be more generic?)
2. FB V Offset would need to change with different *TV settings
3. FB V Offset would need to change if V Offset is changed
 
Maybe it is wrong to mix this in with the other geometry things.

IanSB:
I intend the palette stuff to work with other machines if possible.
The FB V Offset is different for modes06 & mode7 so it really belongs in either the geometry or sampling menus which both have dual settings. My first thought mentioned earlier was to put it in the sampling menu and call it "V delay" to match the existing H delay option but as it interacted with geometry settings I thought it better to put it there.
The frame buffer does need to be bigger than the video to capture any border colour on other systems and now you can reprogram palette colour 0 you can effectively set a border colour on the beeb / electron as well.
E.g I might want to set the frame buffer to 800x300 so that it completely fills my 1600x1200 monitor and when palette 0 is reprogrammed it would cover the whole screen. 
With the new  FB height x2 setting, V height and FB height should always be the same so V height could be dropped.

IanSB:
If we can determine a way of setting the value automatically during calibration, perhaps it should go in the sampling menu after all.

hoglet67:
OK, that makes sense Ian.

I've just been thinking about the CPLD for a faster PSYNC, and your suggestion of making it asymmetric. i.e. 3 shorts and a long, transferring a total of 8 6-bit pixels (which would actually by 2 memory words, is that right?)

One questions: would that not then cause issues with H Offset? If you shifted things by say one PSYNC edge using H Offset, then the longer PSync would no longer line up with a memory word. The conclusion being you could only safely increment H Offset in multiple of 4 (or is it 2?)

Not sure if this is an issue or not...

IanSB:
Ok then then, I'll add V delay to the sampling menu, and FB Height x2 to the geometry menu. We can look at removing V height after that.

> The conclusion being you could only safely increment H Offset in multiple of 4 (or is it 2?)

Perhaps the H delay could be used in some way to provide fine adjustment for H offset.
I think we should try the standard timing approach first as we only need to look at the asymmetric one if that doesn't work.

hoglet67:
Ian,

I've just tried an experiment with doubling the rate of PSYNC (and changing nothing else).

To do this, make the following change to the CPLD:
```
                if counter(4 downto 0) = "00000" then
                    quad(11) <= shift_B(3);
                    quad(10) <= shift_G(3);
...
                    quad(1)  <= shift_G(0);
                    quad(0)  <= shift_R(0);
                    psync    <= counter(5);
                end if;
```
to
```
                if counter(4 downto 0) = "00000" then
                    quad(11) <= shift_B(3);
                    quad(10) <= shift_G(3);
...
                    quad(1)  <= shift_G(0);
                    quad(0)  <= shift_R(0);
                end if;
                psync    <= counter(4);
```
I've verified with a scope that the PSYNC interval decreases from 250ns to 125ns (each quad is sent twice).

Unfortunately, I don't think the software is close to keeping up: either in 4bpp ot 8bpp modes. (I'm testing in Mode 6). The display is very shimmery!

Did you have a more optimised 4bpp copy loop?

Dave

IanSB:
> To do this, make the following change to the CPLD:

I actually tried that a couple of days ago when I was testing the Vsync change and I couldn't get the Pi to boot at all but it looks like I was on the right track. Can you attach your jedec so I can try that.

> Did you have a more optimised 4bpp copy loop?

Have you changed the current loop to drop every other word?

hoglet67:
> I actually tried that a couple of days ago when I was testing the Vsync change and I couldn't get the Pi to boot at all but it looks like I was on the right track. Can you attach your jedec so I can try that.

Sounds like you didn't move the assignment out of the if statement. That would result in PSYNC being stuck low all the time.

Anyway, jedec attached:
[RGBtoHDMI.zip](https://github.com/hoglet67/RGBtoHDMI/files/2943043/RGBtoHDMI.zip)

> Have you changed the current loop to drop every other word?

No, I didn't make any changes. Maybe it wasn't surprising that it didn't work.

IanSB:
> Sounds like you didn't move the assignment out of the if statement. That would result in PSYNC being stuck low all the time.

That was it.

I've added a PSYNC x2 sampling option. It works in 8 bits/pixel mode but 4 bits/pixel is messed up although it's still good enough to check for horizontal glitches.
Initially it glitched as bad as the 192 Mhz test I did but I fixed that by moving the "SKIP PSYNC" code into each capture file.

With BBC timings and a 270 high frame buffer it just about works in 4 and 8 pixel modes and is about 99% glitch free so it's right on the ragged edge (it also works the same with 192Mhz clock!)
However, with CGA timings which use a 14.3 Mhz pixel clock it is stable and works with 4 & 8 bits and double and single height frame buffer so it's definitely worth pursuing.
I've checked in the code so you can try it yourself but ideally you need a source with an NTSC crystal like an Atom with 60Hz TTL RGB to test it. I'm using the following geometry:
geometry06=12,22,83,220,672,440,4,85909091,5472,5000,6

Is there any chance this could made switchable in the current CPLD?
(I think there are 1 or 2 bits that could be removed from the delay setting to use for control)

hoglet67:
> Is there any chance this could made switchable in the current CPLD?
> (I think there are 1 or 2 bits that could be removed from the delay setting to use for control)

This all sounds great.

I have a pair of old IBM 5150's in the loft. I need to check what graphics cards they have. I have a feeling it's the Hercules card (with a 16MHz Pixel clock). But it's possible I'm mis-remembering.

Regarding the CPLD, it's currently 100% full (according to the [fitting notes](https://github.com/hoglet67/RGBtoHDMI/blob/dev/vhdl/fitting.notes)) but I think we can free up some space. We could reduce delay, but I'd like to try something else first.

I'd like to switch to using the rising edge of the clock, and get rid of the fixed delay before we start capture. Doling this will allow the counter to be reduced from 12 bits to 8 bits, freeing up 4 registers. It will also fix the issue we observed with using the 39MHz pixel clock on the ZX81, and possibly we could then drop the half rate sampling modes.

This should allow us to add PSYNC rate to the CPLD scan chain. This would no longer be backwards compatible, so would need to be version 3.x, but that's fine by me. This change will break the smooth horizontal scrolling, but I think that's fixable in software.

Are you OK if I merge your commits back into dev, so I don't end up making conflicting changes doing this?

Dave

IanSB:
> Are you OK if I merge your commits back into dev, so I don't end up making conflicting changes doing this?

Yes it's OK to merge.

> It will also fix the issue we observed with using the 39MHz pixel clock on the ZX81

Using half rate would ensure the 6 bits worked with plenty of headroom which might open up the possibility of doing something with other non-TTL colour computers using some sort of generic pre-processor to quantise the source. At the very least a compatible Atom converter could be made with just the analogue part of your circuit.
I wonder if it could be made to work at 16 Mhz by slightly overclocking the Pi's memory?

hoglet67:
OK, just done a bit more testing prior to merging.

Commit bc103f61 (Optimise skip psync) seems to result in a hang in Mode 7.

Can you take a quick look at that?

IanSB:
> Can you take a quick look at that?

OK

IanSB:
That's done

hoglet67:
OK, all merged now and pushed.

I've also pushed my work so far on adding a programmable rate parameter to the CPLD. This effectively switches the CPLD to a 6 bits/pixel sampling mode, where both the RGB0 and RGB1 inputs are sample in parallel. In this model, PSYNC is toggled every two pixels.

I've updated the cpld_normal.c file to expose this parameter in the sampling menu so you can experiment with it. The CPLD version is now 3.0, so I the system should still work fine against a 2.x CPLD.

At the moment, the setting is independent of your PSYNC x2 sampling setting, but clearly the need to be linked somehow. Lets see if it actually works first.

It may be that the pixel ordering needs tweaking. The results on Mode 6 are not quite as I expected.

Dave

IanSB:
> I've updated the cpld_normal.c file to expose this parameter in the sampling menu so you can experiment with it.

OK thanks.
I've been thinking again about the asymmetric clock and it could be simplified to 1 short and 1 long which would give increased time for the data write and might overcome the marginal timing at 16Mhz.
I drew a timing diagram to illustrate:
timing

The top pair is the old 3 bit timing and the bottom pair is the proposed asymmetric timing.
As far as I can see, this wouldn't require any extra registers to buffer the pixel data and the data would still be latched with a symmetric clock, it's just the PSYNC that would be asymmetric.

hoglet67:
I'd like to verify the current version working first, albeit at a lower max clock rate.

I'm testing with an Atom at moment, and there are no cache misses, but something is definitely not right. I'm not sure if it's the new CPLD, or a but in the PSYNC x2 capture code.

Part of the difficulty was that Phill's board was only stable before if I used the Subsample Even option to take a single sample in the middle of the Atom pixel. If I used Normal, and sampled each pixel twice, I got flickering pixels. So this is making somewhat hard to test.

IanSB:
Yes, I agree to get it working first with new CPLD
The PSYNC code is not correct for 6 bit RGB it just drops 1 word, the bits still have to be moved around

hoglet67:
OK, I think I see the issue(s): there are issues in both the CPLD and the PSYNCx2 Capture Code.

When rate = '0' the 12 data bits are:
```
B3 G3 R3 B2 G2 R2 B1 G1 R1 B0 G0 R0
```
Where the number indicates is the pixel number.
When rate = 1 then you get two quads that are currently:
```
B1' G1' R1' B0' G0' R0' B1 G1 R1 B0 G0 R0
B3' G3' R3' B2' G2' R2' B3 G3 R3 B2 G2 R2
```
(where ' indicate the addition 3 bits/pixel)

I think it would actually be more logical to pack the pixels as follows:
```
B1' G1' R1' B1 G1 R1 B0' G0' R0'  B0 G0 R0
B3' G3' R3' B3 G3 R3 B2' G2' R2' B2 G2 R2
```
i..e the bits for a particular pixel are packed together.

This change is trival and almost free. So I've just pushed it (and a new .jed file).

But as you say, the current capture PSYNC x2 code is not combining the bits correctly. It's simply dropping the second quad.

Do you want to have a go at updating the PSYNC x2 capture code?

IanSB:
> Do you want to have a go at updating the PSYNC x2 capture code?
Looking at that now

PR #31, opened 2019-03-08
Psync x2 working

IanSB:
4 & 8 bits working in Psyncx2 sampling mode.
Strangely 4 bits is more glitchy than 8 bits although it would normally be used in 8 bits anyway.
The type of glitch looks like it is missing the first psync as they look the same as the glitches I fixed previously by moving skip psync into the sampling code. It may be that after moving to the other edge of hsync that psync is now starting too early so could it's start be delayed a bit to see if that makes any difference?.
At the moment the code does this:
Wait for hsync
Setup registers
Jump to selected sampling code
skip psyncs
capture samples

If the Psyncs start too early they may occur during "Setup registers" & "Jump to selected sampling code" sometimes resulting in glitches


hoglet67:
Great, that is working nicely and is now merged.

I've just been playing with the asymmetric psync, and the results are interesting. I've committed the VHDL in case you want to play. You'll need to recompile it if you want to play with the duty cycle.

The falling edge of PSYNC always happens at count = 0.

The rising edge of PSYNC happens at the count set in this line:
```
                if counter(4 downto 0) = 0 or (rate = '1' and counter(4 downto 0) = 16) then
```

The value of 16 will give 50:50 duty cycle.

If the CPLD divisor is 6, then the following count values are skipped:
6,7,14,15,22,23,30,31
i.e. basically xx11x

What I'm finding is that if you move the rising in either direction, things quickly degrade:

8 - completely unstable
10 - many misses
12 - occasional misses
13 - stable
16 - stable
17 - stable
18 - occasional misses
20 - many misses
24 - completely unstable

This was in 8bpp mode.

I'm about to recheck just to confirm I'v not made a mistake here.

But this would suggest that the memory write is actually well pipelined, and that the two halves of the capture are fairly well matched.

If you want this to run at 16MHz, overlocking the memory is certainly one answer.

But another thing to try is a custom version of WAIT_FOR_PSYNC_EDGE that excludes the second I/O read. I certainly needed that on the breadboard, but I never rechecked on the final PCB whether noise on PSYNC was an issue.

Dave

IanSB:
I just moved a lot of the setup registers from after hsync to before and that got rid of virtually all the glitches and it now only glitches once every few seconds so maybe the asymmetric will mop up those last blips (Just trying to figure out how I broke mode 7 again!)
i.e.
Setup registers
Wait for hsync
Jump to selected sampling code
skip psyncs
capture samples

IanSB:
Based on previous experience I think that the blx r10 is the issue so I'm going to move HSYNC detection into each sample loop as well so that takes the blx out of the time sensitive part.
i.e.
Setup registers
Jump to selected sampling code
Wait for hsync
skip psyncs
capture samples

IanSB:
The new CPLD appears to have broken mode 7

hoglet67:
OK, I'll look at that next.

hoglet67:
Hmm, Mode 7 seems OK to me.

We need to increase the default H Offset from 24 to 32, but other than that is seems fine.

What problem are you seeing?

IanSB:
No deinterlacing and other wierdness, I think it might be software after all.

hoglet67:
OK, let me know if I can help.

I've just tried commenting out these lines in the WAIT_FOR_PSYNC_EDGE macro:
```
        // Check again in case of noise
        ldr    r8, [r4]
        eor    r8, r3
        tst    r8, #PSYNC_MASK
        bne    wait\@
```
It then runs nicely at 16MHz with Rate=On in the sampling menu, and it does so in 4bpp and 8bpp, and in all sampling modes.

I think this is buying us a huge amount of margin.

There are some minor artefact, but these were fixed by skewing PSYNC with respect to the data changing. No asymmetric PSYNC is needed.

On reflection, if the data and psync change at the same time, then there is a race condition. The double read of PSYNC was effectively working around this, but is very expensive. Adding some skew in the CPLD is a much better solution.

Dave

IanSB:
Is that skew in the current CPLD or are you going to check that in?

hoglet67:
I've just pushed a version with one cycle of skew.

That seems to be enough for me, but it might be work adding a couple more to be safe.

hoglet67:
(I should also check my Atom Pi Adapter board, it may behave differently)

PR #32, opened 2019-03-09
Integrate six bit sampling support (and other related work)

IanSB:
Six bit sampling support integrated and new table driven mechanism for selecting the capture loop.
Six bit looks very stable now at 16Mhz so may well work with EGA.
Note H offset is automatically compensated when six bit switched on so there is no visible difference in the screen but things like in band and the odd/even/double pixel sampling won't work as there are no capture loops yet.
Alignment doesn't work in 8 bits/pixel mode as it assumes the buffer is 4 bits.

### DIFF
--- a/src/capture_line_half_4bpp_8bpp.S
+++ b/src/capture_line_half_4bpp_8bpp.S
@@ -23,153 +23,152 @@
         b       preload_capture_line_half_4bpp
 capture_line_half_4bpp:
         push    {lr}
-        cmp     r1, #400/8               //sanity check on buffer size as only capturing half of pixels so width >400 will never finish   
-        movgt   r1, r1, lsr#1
-        cmp     r5, #DUPLICATE_HEIGHT
-        orrle   r3, r3, #BITDUP_NO_LINE_DOUBLE
-        bicgt   r3, r3, #BITDUP_NO_LINE_DOUBLE 
-        SKIP_PSYNC
-capture_half_4bpp:                          
-        WAIT_FOR_PSYNC_EDGE              // expects GPLEV0 in r4, result in r8
-        CAPTURE_LOW_BITS                 // input in r8, result in r10, corrupts r9
-        WAIT_FOR_PSYNC_EDGE              // expects GPLEV0 in r4, result in r8
-        CAPTURE_HIGH_BITS                // input in r8, result in r9/r10
-        mov     r11, r10                 // save first word
-        WAIT_FOR_PSYNC_EDGE              // expects GPLEV0 in r4, result in r8
-        CAPTURE_LOW_BITS                 // input in r8, result in r10, corrupts r9
-        WAIT_FOR_PSYNC_EDGE              // expects GPLEV0 in r4, result in r8
-        CAPTURE_HIGH_BITS                // input in r8, result in r9/r10
-        
-        tst    r3, #BIT_EVEN_SAMPLES
-        andne  r9, r11, #0x00000007 
-        movne  r12, r9, lsl#4
-        andne  r9, r11, #0x00000700  
-        orrne  r12, r12, r9, lsr #8
-        andne  r9, r11, #0x00070000 
-        orrne  r12, r12, r9, lsr #4
-        andne  r9, r11, #0x07000000 
-        orrne  r12, r12, r9, lsr #16             
-
-        andne  r9, r10, #0x00000007 
-        orrne  r12, r12, r9, lsl#20
-        andne  r9, r10, #0x00000700  
-        orrne  r12, r12, r9, lsl #8
-        andne  r9, r10, #0x00070000 
-        orrne  r12, r12, r9, lsl #12
-        andne  r9, r10, #0x07000000 
-        orrne  r10, r12, r9      
-       
-        tst    r3, #BIT_ODD_SAMPLES
-        andne  r9, r11, #0x00000070 
-        movne  r12, r9
-        andne  r9, r11, #0x00007000  
-        orrne  r12, r12, r9, lsr #12
-        andne  r9, r11, #0x00700000 
-        orrne  r12, r12, r9, lsr #8
-        andne  r9, r11, #0x70000000 
-        orrne  r12, r12, r9, lsr #20             
-
-        andne  r9, r10, #0x00000070 
-        orrne  r12, r12, r9, lsl#16
-        andne  r9, r10, #0x00007000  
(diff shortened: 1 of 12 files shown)


hoglet67:
Hi Ian,

I've merged this, and made a couple of changes.

1. Removed trailing white space from .S files

2. Added PSYNC skew into Atom CPLD (again, 1 cycle was sufficient, I added 2 to be safe)

3. Separated cpld->update_cap_info() from cpld->set_mode(). This allowed me to make the Normal CPLD parameters private to the Normal CPLD again. You had needed to expose them to directly access the SIXBITS parameter. I've always wanted CPLD parameters as opaque to the rest of the system. The Atom CPLD doesn't have a SIXBITS parameter. It currently works, because the default behaviour is to return 0 for an known parameter. But we should be relying on this. I think what I have done is functional the same, and the H Offset scaling magic happens when selecting SIXBIT mode.

One issue we have to fix is that we have broken backwards compatibility with CPLDv1 and CPLDv2 users. Specifically, removing the second IO read in WAIT_FOR_PSYNC_EDGE in all cases, on reflection, is not right. It should really only be removed for CPLDv3 users. I've tested this and do see some noise in mode 0 with CPLDv2 (withing the PSYNC skew). Couple of possible solutions:
- yet another r3 bit to make these conditional
- some parameterized macro magic to create two versions of each of the line capture functions

I'll have a try at the first of these, and see if that affects the 8MHz

I'm there's a slightly strange interaction between the delay value in Sampling=Double mode that suggests something is amiss somewhere. Even values of delay are fine, Odd values are not as I would expect. I'll see if this was present in older code.

Anyway, this is all looking very nice.

What's next on your stack?

hoglet67:
I've made the second read of PSYNC conditional on sample width, so it's only suppressed in PSYNC x2 mode. This seems to get the best of both worlds: works with CPLDv2 and doesn't seem to disrupt the faster capture stuff. Commit 60818274.

hoglet67:
There is one small UI glitch I've just spotted...

 If Six bits = On and FB Bits/pixel  = 4 then the palette is slightly messed up when the OSD comes up. This is because capture_line_sixbits_4bpp captures 4 bit pixels.

Is this something you are actually planning to make use of?

I had assumed all the wider pixel modes would use the 8bpp frame buffer,

hoglet67:
I'm going to start adding issues for bugs I find.

Could you do the same if there is anything you are aware of please?

IanSB:
> If Six bits = On and FB Bits/pixel = 4 then the palette is slightly messed up ... Is this something you are actually planning to make use of?

I'm aware of that, I put it in in case there was insufficient time to do things line NTSC artifacting at 8bpp

> What's next on your stack?

Wiring up 4 bit to CGA and get that working to test >3 bits actually works (I don't have an EGA card at the moment but may get one off ebay)
Then go back to the FB heightx2 and V delay menu additions and the rest of the in band command code

One thought:
Now that 6 bits are available it opens up the possibility that this could be made to work with any pre 16 bit computer using an appropriate front end processing circuit to correctly quantise the source. E.g. your atom one or my zx81 one.
It gets a bit more complicated with something like a commodore 64 which generates s-video within the VIC II so there is no digital source available. This could be processed using one of those front end TV chips to pal/ntsc decode but it would also need some fast front end single chip processor (Atmel maybe?) to analyse the decoded 24 bit rgb data and quantise that back to the original 4 bits.
Something like that could be used as a generic solution but would need to communicate with the Pi to switch modes etc so is there anything that could be added to the CPLD at the moment to support communication / control of any proposed front end processor?
e.g.
Looking at the VHDL there is the following:
NET "spare" LOC = "P19"; # input gpio22
So there is a spare CPLD pin and GPIO
If the GPIO was defined as an input, could the CPLD pin could be defined as an output with a clamp signal for the atom circuit? 
Are there any other unused GPIOs? If not then the above track could be cut and the GPIO used as an SPI select line?

hoglet67:
I was looking at pin counts the other day with a view to increasing the pixel bus width from 12 to 16 bits. But I think the double rate capture eliminates the need to do this.

In a new board design for 6-bit RGB capture we can free up a few GPIOs on the Pi:
- SPARE (gpio22)
- MUX (gpio24) 
- LED (gpio27)
- Serial RX (gpio15) (we just need Serial TX for debugging)
- MODE7 (gpio25) (which selects between /6 and /8) could be moved into the CPLD scan chain
- and possibly SP_CLKEN_PIN (gpio1) (the clock enable for the scan chain, which I added due to noise in the prototype

It would be very nice to design a general purpose front end, that used the thresholded comparators like the Atom board.

If you want to hack something with the current board, then I'm sure that would be possible. I'm slightly confused why you want to pass any control signals through the CPLD. Why not just connect the processor to the Pi directly.

In terms of verifying the 6-bit capture, I've just added a example to cmdline.txt that uses the this on an Atom to give us back orange without any specialized hardware.

The pixel format is B1 G1 R1 B0 G0 R0.

It turns out that when orange is output, the green is half intensity, and the CPLD and 74LS08 thresholds are sufficiently different that G1 =1 and G0 = 0. So with a suitable palette that ignores B1 and R1 you can get orange to display nicely. It's a bit of a hack, but it shows the 6-bit mode works. The pixel clock is quite slow though: 7.159MHz. So it's not pushing the system.

IanSB:
> It would be very nice to design a general purpose front end, that used the thresholded comparators like the Atom board.

Yes, with three sets of comparators for RGB or YUV you could get support for monochrome, Atom / TRS80 colour computer, Amstrad CPC464 (uses 3 levels each of RGB), Spectrum and probably others.

>  I'm slightly confused why you want to pass any control signals through the CPLD. Why not just connect the processor to the Pi directly.

I think the only thing that could be added to the CPLD would be the clamp signal any comms would be directly from GPIO.
Perhaps a generic comparator solution should use DACs to set the thresholds so no need to tweak pots and maybe it could even auto calibrate with an appropriate test pattern.

IanSB:
I've just been testing mode7 at 192 MHz clock rate with a new mode 7 half odd/even sample loop and it seems to be completely stable. Although this is well out of spec it is potentially useful for those beebs with a very asymmetric 12 Mhz clock as it doubles the resolution and it removed the last twinking pixels on one of my systems. Unfortunately with double the psync rate there is no time to read memory and so deinterlacing doesn't work. 
Could be implemented in the CPLD with a switch that enabled half odd/even mode? (i.e. put every other pixel in the shift reg and halve psync)
This would work with the existing capture loop without any changes.

PR #52, opened 2019-03-11
Rework m7disable and Elk settings

IanSB:
I've tried to genericise these settings a bit:
m7disable is now Auto Switching with values of: Off , BBC Mode 7 , PC CGA/EGA
This setting determines if the system will auto switch settings depending on the source mode
PC EGA outputs will switch back to CGA when those modes are selected so would need 2 sets of timings like mode06 & 7
(PC CGA/EGA is a placeholder but I will have an EGA card shortly to test)

Elk is now V Sync Type with settings of Standard & Alt (Electron)
This is just a cosmetic change and it is still referred in the code as elk but will allow us to add any further handlers for problem vsync.
Is the Electron Vsync unusual or is this likely to be of use by other machines?
The command line settings have changed to "autoswitch" and "vsynctype"


hoglet67:
> Is the Electron Vsync unusual or is this likely to be of use by other machines?

The difference seems to be where the active video starts in each field. Specifically,  in the second field (the even field) , the active video in the Beeb starts one line earlier compared to the Electron. 

If this is not taken into account, then when the two fields are overlaid on top of each other, the screen appears to judder up and down by one line. So what the capture software does to fix this in the Beeb case is to reduce V Offset by one for the even field.

Strictly speaking, I'm not sure whether it's the Beeb or the Electron that's at fault here, but they are different by one line. (This only affects the interlaced case).

Now it has just occurred to me that there is a small difference in VSync between the Beeb and the Electron. In the Beeb case, inverted sync pulses are present during Vsync.  In the Electron case they are not. It's possible, I suppose, this is affecting where we think VSync ends, and hence where we start counting lines in the field.

An interesting experiment on the Beeb would be to replace the XOR gate generating CSYNC with an OR gate, and see if that changes how we decode the fields. I might have a go at that now...

Dave

hoglet67:
> Now it has just occurred to me that there is a small difference in VSync between the Beeb and the Electron. In the Beeb case, inverted sync pulses are present during Vsync. In the Electron case they are not. It's possible, I suppose, this is affecting where we think VSync ends, and hence where we start counting lines in the field.
>
> An interesting experiment on the Beeb would be to replace the XOR gate generating CSYNC with an OR gate, and see if that changes how we decode the fields. I might have a go at that now...

For get that bit, it's complete rubbish. The Beeb already uses a NOR gate, so both the Elk and the Beeb have an identical VSYNC. Only the Master includes the inverted HSYNC pulses.

PR #56, opened 2019-03-14
Full support for separate and composite syncs also delay rework

IanSB:
I've modified the CPLD v4 to fully support separate syncs and changed the sofware to support this.
I've added an invert bit to the CPLD to invert the CSYNC line and fed the link input straight to the spare output so the Pi can look at VSYNCS (using CSYNC as HSYNC)
The Pi auto detected composite or separate syncs and automatically inverts them as appropriate so the existing sync detection still works.  I think I've tested all combination of sync polarities with different sources.
Unfortunately to fit that in I had to remove the extra 2 delay bits so delay is now only 2 bits long with the extra 2 bits made up by adjusting the h offset. This seems to work OK for manual adjustment but auto configuration of the delay in mode 0 doesn't work (It still works in mode 7)


hoglet67:
Ian,

I think this one might need a bit more discussion...

I'm wondering if this has turned out much complex that it needed to. Did you really convince yourself that the only solution was to feed vsync through on a separate signal and make this visible to the software? I'd like to understand the reasoning here.

I'm probably missing something, but here's my thinking...

My gut feeling is that as long the CPLD generates csync = hsync xnor vsync, the software should be able to work with that.  

There are only two possible shapes for this combined csync signal:
- if hsync and vsync are the same polarity, then we have to deal with a normal composite sync signal
- if hsync and vsync are opposite polarities, then we have  deal with an inverted composite sync signal

All an analyze_sync function needs to do is to measure the average value of csync to determine whether it's normal or inverted. All the software then needs to do is eor the signal with a bit that says the sync was inverted.

i.e. I don't see why the software needs to ever see the original vsync, or care whether it was inverted or not.

If we could possibly avoid complicating the physical interface between the CPLD and the Pi, I would be really keen to fully explore that option first.  Bringing an extra signal into play here (the spare) signal should be our last resort.

A couple smaller points on the CPLD change, if we do decide to proceed.

1. If the scan register format changes, we should  (in theory) update the version number. I've been doing that quite strictly, even during development, so that I'm sure the latest software is compatible with every past version of the CPLD. This is mostly a safety net: if we start having a weird issues with the latest CPLD we easily can work back through all previous versions.

2. If you reduce the size of delay, you can also reduce the size of counter. So a 3-bit delay would probably have fitted.

Anyway, I'll wait for your thoughts on my main point before going any further with merging this.

Dave

hoglet67:
By the way, I mean to ask earlier. What git commands are you using to sync your branch?

I'm seeing a lot of empty merge commits, where I would have expected your branch would have just been able to be fast forwarded. These make the history look less non-linear that it needs to.

hoglet67:
I've been thinking about the sync inversion stuff a bit more.

My main motivation here is to use the minimal CPLD resources, so as we leave as much scope for future enhancements as possible.

The scheme I outlined above doesn't quite work, because the CPLD also (obviously) still needs to know if the sync is inverted.

So I think the miminal change here is:
- increment the CPLD version to 5.x
- add the invert bit to the scan chain
- csync1 <= (S nxor link) xor invert;

The analyze_sync function would initially set invert to 0, measure the average value of csync over a period of time like 20ms, and if that's closer to 0 than to 1, then invert would be set.
 
And that's it. No other software changes required. And the software doesn't even need to see vsync.

That change uses the one spare macro cell (for the additional register) and two extra product terms.

For now we can stick with the 4-bit delay for now (which is easier to use from a user point of view). If in the future we need to free up resources, both delay and counter can be reduced by either 1 or 2 bits, freeing up 2 or 4 macro cells respectively.

Again, I'm possibly missing something here...

Dave

IanSB:
The reason for the separate syncs was that the sync polarities were originally used to signal different modes/resolutions so I was intending to use that to do the auto switching. However it could probably be done by measuring the H & V timing instead. The sync invert bit as you mention above is more essential as EGA would result in an inverted CSYNC signal.
I'm using a windows Git client (SmartGit).

hoglet67:
OK, so what should we do? I think I would prefer to avoid sending vsync across.

I've got the changes locally for CPLDv5 and the corresponding cpld_normal.c class.

Sync can be set manually in the config file, and the system will work with inverted sync.

I just haven't merged in your analyse_sync. Without this, incorrect sync polarity causes the capture loop to hang.

Did I notice some other changes in the pull request, to do with WAIT_FOR_PSYNC_EDGE_FAST? Are these related to the sync polarity? Or is that a separate optimization.

Dave

IanSB:
Sorry forgot to mention the FAST change, that was to do with getting 640x480 VGA working. It also needs a new fast capture loop as well which I haven't done yet (I was using a tweaked version of the standard one which I reverted before committing).
I think it should be possible to detect the mode changes by the H and V timing changes so we can go with just the sync invert bit for now and keep the Vsync/delay part in reserve if there are issues with mode detection going forward.

hoglet67:
Sorry this process is getting slightly messy.

I've just pushed my local changes to a cpld_v5 branch (this is just the CPLD and cpld_normal).

I'm just about to try to add analyse_sync, so hang fire for a bit.

hoglet67:
This is now implemented as discussed (based on just looking at csync) and merged back to dev.

Best to consider the cpld_v4 and cpld_v5 branches as having served there purpose, and any future work should be done in dev.

I've just checked, and when you sync your local dev branch, it should be fast forward merge.

These are the commands I would use.
- git checkout dev
- git fetch upstream
- gt merge upstream/dev

How have you been doing it?

I've also manually merged the the other changes relating for _FAST. Do check commit 0a9b31ff, and make sure I've not included anything I shouldn't have.

There are some other small bits of work I've also just pushed to dev. They should be self-explanatory if you work through the commits.

Right, I think that's me done now for a week or so.

Have fun!

Dave

hoglet67:
Everything relevant now merged I think.

It's all still in Ian's cpld_v4 branch if we ever need to pull in the delay related stuff.

hoglet67:
One more thing...

One TODO item with what I have pushed is that it doesn't yet cope with a change in sync polarity mid stream. 

Did your version manage to cope with that case, i.e. did a change in sync polarity result in a mode change being flagged?

If the sync polarity is incorrect, what currently what happens is that wait_for_vsync() in the capture loop hangs. I think waiting for a "short" low-going sync pulse which it never sees.

Unfortunately that locks up the UI, because key-presses are not noticed.

I wonder if wait_for_vsync could time out after say 25ms. That could then trigger a mode change, and the cpld->analyse() would get called again and would set the invert bit correctly.

The timeout could also blank the screen (as you suggested in #7)

IanSB:
> Did your version manage to cope with that case, i.e. did a change in sync polarity result in a mode change being flagged?

No, not yet.

> I've just checked, and when you sync your local dev branch, it should be fast forward merge. How have you been doing it?

It's got a GUI and I've been using the wrong option.

hoglet67:
> It's got a GUI and I've been using the wrong option.

What tool are you using?

IanSB:
SmartGit

Issue #34, opened 2019-03-09
Bug: Smooth Horizontal Scrolling is broken


hoglet67:
Due to the change in sync edge used.

hoglet67:
The test for this is Tricky's RallyX

Once started, you need to press U S B

Then use the cursor keys.

For CPLDv1 and CPLDv2 we use the following code:
```
        // Start with the configured horizontal offset
        // Implement half character horizontal scrolling:
        // - a "short"  hsync is 3.5us, leave h_offset as-is
        // - a "normal" hsync is 4.0us, increment h_offset by 1
        // - a "long"   hsync is 4.5us, increment h_offset by 2
        // So test against two thresholds inbetween these values
        cmp    r10, #(4000 + 224)
        addgt  r7, r7, #1
        cmp    r10, #(4000 - 224)
        addgt  r7, r7, #1
```

For CPLDv3 this needs changing to:
```
        // Start with the configured horizontal offset
        // Implement half character horizontal scrolling:
        // - a "short"  hsync is 3.5us, increment h_offset by 2
        // - a "normal" hsync is 4.0us, increment h_offset by 1
        // - a "long"   hsync is 4.5us, leave h_offset as-is
        // So test against two thresholds inbetween these values
        cmp    r10, #(4000 + 224)
        addlt  r7, r7, #1
        cmp    r10, #(4000 - 224)
        addlt  r7, r7, #1
```

Only problem is, this is not backwards compatible.

So yet another flag...

I was trying to avoid a CPLDv3 flag, but maybe this is easiest....

hoglet67:
Closing as fixed.

PR #58, opened 2019-03-15
Optimise capture loops and fix horizontal smooth scrolling

IanSB:
Optimise the capture loops even more so that VGA text mode works.
This has a pixel clock of 28.322 MHz so even faster than VGA graphics mode
Also fix V3: Smooth Horizontal Scrolling is broken #34
capture33


IanSB:
I've added a profile system using one file per profile each of which contains the usual command line options but they can be on multiple lines. These should be copied to a "Profiles" folder in the root of the SD card.
A "Default" file contains all the default settings and each profile file contains settings that differ from the default so changes such as the key mapping only need to be made to the default file.
There is a new menu option "Computer Profile" which selects the different profiles and the current setting is written to cmdline.txt as "profile=x" where x is the profile number so the profile will persist across a reset/power cycle. (At the moment the profiles themselves still have to be manually edited - only put command line settings in the profile text files)

hoglet67:
The profile stuff looks very nice; thanks for making at start on this.

I can think of a couple of further things that might improve it  (probably you have these in mind as well)..

1. Persisting the auto calibration to the currently selected profile would be very cool.

2. Rather than have us (as developers) statically decide on the set of profiles (in osd.c) at compile time, it would be nice to build this list dynamically at boot time from the files in the profiles directory. I guess we would need to add a "profile_name" property into the profile file (or maybe fall back to the filename if this is missing). Then people can add their own profiles easily without needing to recompile.  They could support multiple instances of the same system (e.g. Beeb 1, Beeb 2), just by copying the profile file and changing the profile name, The profile=x in cmdline.x might become the filename of the current profile.

3. Somehow I would like to get away from the comma separated parameters lists that we have for sampling and geometry, as these are fragile and a pain if you just want to change one parameter, but it's at the end of the list. This will need  some quite careful thought. The geometry properties are essentially fixed, but the sampling properties depend on the CPLD Implementation (i.e. the Atom CPLD only has a three parameters).  Also, somehow we need to allow each property to appear multiple times (the main use case being systems that support multiple different modes, like the Beeb).  So I was thinking the files could support multiple Mode sections. I think I would prefer this, rather than two separate profile files, as it just seems nice to have it all in one file. But I can think of a huge number of nasty corner cases, so I would suggest we don't rush into this without fully thinking it through.

(3) is something I'd quite like to work on when I'm back. But feel free to dive into (1) or (2). Or of course, anything else that takes your fancy.

IanSB:
> Persisting the auto calibration to the currently selected profile would be very cool.

Yes but it needs to be an explicit user choice, not automatic as I've found that better calibration can be made with specific test images especially in mode 7 and I wouldn't want to lose that by accidentally hitting the calibrate button.

>  it would be nice to build this list dynamically at boot time from the files in the profiles directory.

Yes I have considered that but at the moment it's not feasible until I have the PC auto switching working.
The initial autoswitch code will have to know which profile to select which is going to be more complex if the user can add/remove/rearrange profiles. Once I know exactly what is required to autoswitch we can look at adding any parameters to the profiles to support that and move to a more flexible system.
Also things like the output mode should be part of the profile because it should be changed when switching from 50Hz to 60Hz sources etc. I think you found the relevant registers in the Pi so we need to add the resoloution and refresh rate to the menus and each profile as required if we can get that working (maybe the scaling kernel as well)

> Also, somehow we need to allow each property to appear multiple times (the main use case being systems that support multiple different modes, like the Beeb

Not sure of the best way to go with that, the PC side has 8 profiles so far with probably more to follow so should they all be put into one "PC" profile or multile "PC xxxx" profiles that get selected as appropriate.
Perhaps a master profile file pointing to each individual profile (e.g. one BBC master profile file pointing to mode0-6 and mode7 profile files and one "PC" master profile file pointing to 8 PC profile files)

hoglet67:
> Yes but it needs to be an explicit user choice

Agreed. 

> Yes I have considered that but at the moment it's not feasible until I have the PC auto switching working.
> The initial autoswitch code will have to know which profile to select which is going to be more complex if the user can add/remove/rearrange profiles. Once I know exactly what is required to autoswitch we can look at adding any parameters to the profiles to support that and move to a more flexible system.

If I'm understanding, you are saying the only code that needs to reference a specific profile is the PC mode auto switch, because each PC mode is a currently separate profile. That makes sense, given where we currently are.  It seems the requirement is to give certain profiles a stable/unique ID so the code can find them. That is currently satisfied by the enum, but that's inflexible. Could we not just add an optional ID attribute to certain profiles (the value could be an integer or a string, which ever makes life easier). At a pinch, the filename itself could be the ID.

> Also things like the output mode should be part of the profile because it should be changed when switching from 50Hz to 60Hz sources etc. I think you found the relevant registers in the Pi so we need to add the resoloution and refresh rate to the menus and each profile as required if we can get that working (maybe the scaling kernel as well)

This would be nice, I agree, but this feels separate to whether or not we dynamically populate the profiles.

> > Also, somehow we need to allow each property to appear multiple times (the main use case being systems that support multiple different modes, like the Beeb
> 
> Not sure of the best way to go with that, the PC side has 8 profiles so far with probably more to follow so should they all be put into one "PC" profile or multiple "PC xxxx" profiles that get selected as appropriate.

I'm not sure either.

The tricky thing, as you point out, is mode auto switching which needs to "know" which profile to select. 

Do you think the mode auto switching is always going to be hard coded  to cover a limited set of specific cases (like the Beeb and the PC)? Or do you see it evolving into something that can untimately by parameterised?

I had hoped that only a few systems would need to use this (like the Beeb) and we would be able to fudge it. But clearly the PC is another valid use case (where people can change mode on a running system).

Are you aware of any other retro systems that are in scope for us, where the screen mode can change in use?

> Perhaps a master profile file pointing to each individual profile (e.g. one BBC master profile file pointing to mode0-6 and mode7 profile files and one "PC" master profile file pointing to 8 PC profile files)
> 
> EDIT: How about single text files in the profile folder are for machines with one profile but a sub-folder in the profile folder means an autoswitching "set" and that sub-folder contains all the auto switching profiles for that set. So there would be a "BBC" sub-folder with 2 files in it and a "PC" sub-folder with 8 files in it etc.

We need to think about this some more. But I'm nervous about multiple levels of profiles like this, because they are harder for a user to comprehend, and harder to manage if as it involved editing multiple files.

IanSB:
I've done some initial analysis on the various PC modes (not looked at MDA / Hercules yet)

Mode                 H period us V lines  Sync    
VGA TEXT           31.7777     449      -H+V (Inverted csync)
EGA on VGA      31.7777     449      +H-V (Inverted csync)
CGA on VGA      31.459      449      -H+V (Inverted csync)
VGA Graphics     31.7777     525      -H-V (Non-inverted csync)
Real EGA             45.764      365      +H-V (Inverted csync)
Real CGA            63.695      262      +H+V (Non-inverted csync)

From the above, it is not possible to detect the difference between VGA TEXT and EGA on VGA without separate H and V syncs. These two modes have different clock speeds and geometries. (There may be some subtle variation in H and V sync periods but I don't think we could rely on that as "compatible" hardware could have different variations.
I have a possible solution to adding vsync to the CPLD without too much change in resource usage:
The mux functionality should be eliminated so that pin can then be used to control sync invert. This means the invert bit can be removed from the register chain and Vsync could then be fed via the spare pin without needing any further changes. (I have tried this and it fits with 1 register spare)
Mux functionality can be re-instated in two possible ways:
1. Software: When mux is required, switch to six bit mode and capture the other 3 bits
2. Hardware: When 3 bit mode is selected, the sync invert line functions as mux. When 6 bit mode is selected the sync invert line functions as normal (This assumes we will never need to support inverted sync in 3 bit mode)

hoglet67:
Out of your two options: I think I prefer (2).  Dropping the hardware mux entirely and relying on 6-bit mode and software to do the selection seems to be shifting quite a lot of complexity onto the software, which is already getting really complex.

Another way we could do this is to use of psync to indicate the polarity of vsync. For example, by outputting vsync on psync when csync1 is active (low). Or alternatively when version is active (low). That should be fit, even without dropping sync invert from the scan chain. 

A third way might be to use the sp_data input for sync invert, when the scan chain is not being programmed. That leaves mux as is.

Is the only part of the software that needs to use Vsync is the auto detection of PC mode? Does Vsync need to be available all the time, or can it just be sampled occasionally during the blanking interval?

Stepping back, this does feel like it's starting to getting a bit messy, because we are trying to fit everything into a single CPLD, and we have no room to manoeuvre because we are right at the resource limit of the CPLD we are using. 

If the goal here is non-acorn systems, some of which need the hardware for 2-bit RGB digitisation, 
I wonder if we should be thinking ahead to a new version of the RGBtoHDMI hat? We could provide a path for VSYNC into the Pi that bypassed the CPLD (some kind of level shifting would still be needed). Or we could choose a completely different (larger) CPLD if that helps. Or we could drop the multiple offsets, as this is highly Beeb specific.

Dave

IanSB:
I've just added a first attempt at auto profile switching which works but is hard coded at the moment so not very flexible. I also changed the clock and line length to be displayed as the actual pixel clock and line length in pixels rather than x6 or x8 as it makes it easier to use published info when setting things up. (The intermediate values have no real meaning as those entries should always be divisible by 6 or 8)

> For example, by outputting vsync on psync when csync1 is active (low). Or alternatively when version is active (low). 

I think either of those would be OK

There may be a slight issue with PSYNC. I found a DOS program that forced my original card into EGA mode and it appeared to have the correct timing and clock but I couldn't get it to capture the last character on each line. The timings for EGA are very tight as it uses a 16.257Mhz clock with a line length of 45.76uS and active period of 39.37us leaving 6.39us for blanking and sync. It's possible the last valid psync edge is being missed because it is too close to the start of h sync or it may be that the card itself is having problems with the timings. I think I'll write a test program that puts the last char on every line and then look at the output with a scope to see if it's really on the video as I don't have any other display that would work with that mode.

I now have another card with TTL output, this new one can select the startup mode with DIP switches and boot up in MDA/CGA/EGA/VGA etc. I was hoping this would help with the above issue but unfortunately the the EGA & CGA modes on this one don't appear to be using standard clocks and it was difficult to setup the profiles for them as the clock and line length were unknown (the board only had a 50Mhz crystal so likely using an internal PLL). I eventually found a way to work it out using the debug output but I think we need some sort of clock setup menu to help out with that.

IanSB:
I've performed the last character test mentioned above and the last char is definitely there but it is sent after the start of the sync pulse!
This is a whole line of text (cyan) with hsync pulse (yellow) where you can see there is no front porch:
text line
This is a closeup of last char only on screen and start of hsync pulse and it is being sent after the start of +ve going hsync!
last char
Not sure if this is actual ega timing or just a quirk of this card but it's no wonder I can't sample the last character! Not sure if this is fixable as I assume the psync stops on that rising edge.

hoglet67:
Looking good....

I think changing Clock/Line Len to relate directly to the pixel clock makes sense. An additional benefit is you can then toggle between /6 and /8 without changing any other settings.

Regarding psync, currently when csync is low the counter is held in reset, so any pixel data during the sync pulse would not be captured. To allow the counter to continue running during csync and reset on the rising edge would cost an extra register, which I don't think we currently have. To free one up we would have to either shorten the delay counter (currently 4 bits) or shorten the the csync deglitching counter (currently 2 bits).

It is worth noting that a large chunk of the CPLD relates to having multiple offsets:
- most of the scan chain (18 bits)
- the index counter (3 bits)
- the current offset register (3 bits)
So if only one offset was needed, then you could save 21 registers (nearly 1/3 of the CPLD). The multiple offsets are only a benefit of the pixel clock is asymmetric, like in Mode 7. This is probably quite Beeb specific.

Dave

IanSB:
> To free one up we would have to either shorten the delay counter (currently 4 bits) or shorten the the csync deglitching counter (currently 2 bits).

I've put in support for a 2 bit hardware delay. There is no change as far as the user is concerned in that the delay can still be adjusted 0-15 but only 2 bits are sent to the hardware and the higher 2 bits are used to adjust h offset. At the moment it's still writing 4 bits to the hardware but the high two bits are masked to 0 so this version should be able to run on any CPLD with 2 or 4 bit delay support once the chain code has been updated. (This can be easily rolled back if we don't go ahead with the 2 bit delay change.)

IanSB:
I've changed the way the core clock and divisor are calculated.
Previously the divisor and core clock were calculated from the nominal pixel clock and the core clock was then adjusted to match the PPM error. This had problems when the core clock was close to 400Mhz as any adjustment over 400 Mhz wouldn't work even with sanity checking removed and it generally limited the range of adjustment.
Now the pixel clock is adjusted by the PPM error before calculating the divisor and core clock which allows the adjusted pixel clock to work over a larger range.
This helps with finding unknown pixel clocks as you can rapidly sweep the effective pixel clock through a large range by adjusting the line length (as long as clock tolerance set to 0). Once no beat patterns are observed on the screen (i.e. all clean or all noisy) then read the clock from the debug log and use that, then you can find the line length by adjusting it back until no beat pattern observed again.
The effective clock speed should be displayed on the info menu so the logs don't have to be used.

IanSB:
I've added the ability to make the frame buffer larger than the capture area, in which case the capture area is put into the centre of the frame buffer.
Also I changed the H offset and H width to be displayed in pixels rather than characters to be consistent with the frame buffer size.
H offset is now adjusted 4 pixels at a time and H width and FB width are now adjusted 16 pixels at a time.
This effectively makes the "extra" frame buffer around the capture area equivalent to the overscan setting except that you can change it's palette which is useful if logical colour 0 is reprogrammed. Previously you got a small border of logical 0 colour and a larger border of black where the overscan was, now you get a full border of logical colour 0.
Compare this to the version posted on stardot (uses an 800x300 buffer with a 672x270 capture area)
800X300

After this change, the frame buffer could be removed entirely from the geometry menu and calculated automatically based on scaling settings and current screen size

IanSB:
I've added automatic scaling with three options:
Integer
Scales up the image to fill the screen as much as possible using integer pixel scaling.
Non Integer
Zooms the screen from the integer size to fill the screen while maintaining the same aspect ratio (requires interpolation).
Note that both of the above ignore the frame buffer size and generate their own frame buffer size to get the scaling correct based on the captured area and the current graphics output mode.
Manual
Uses the frame buffer sizes in the geometry menu. Adjust the FB sizes to get any arbitrary scaling

There is now no need for the overscan settings and they should be removed or set to zero although you can still use them in conjunction with manual scaling mode to get the old behaviour if required.

I think the menus need to be reworked as there are now a large number of options in the geometry menu, perhaps we need to split off some into a clock menu or similar.

IanSB:
Here's a framegrab with the new palette:
capture32

hoglet67:
Just merged this back into dev.

A few very minor things...

1. Missing Profiles - some kind of on screen error would be nice if the profile could not be loaded. I initially forgot to copy over the Profiles directory.

2. Hard-coded defaults (when no profile loaded) should work for the Beeb

3. The default for the position of "Return" in the menus seems to have changed.

4. The default for the key map seems to have changed.

IanSB:
> The default for the position of "Return" in the menus seems to have changed. The default for the key map seems to have changed.

Edit the default.txt file in profiles to remove those settings and it will revert to your standard ones

BTW the addition of the larger frame buffers has broken mode 7 alignment (and probably mode0 but that's already broken with 8 bit buffers)
Mode 7 alignment works in manual mode when the default frame buffer size is used but in integer and non-integer where it auto creates larger buffers than the capture are it doesn't work.

IanSB:
Could you have a look at updating the CPLD to add the things discussed above:
1. Remove 2 msbs of delay
2. Make psync run during hsync
3. Make Vsync appear on psync when version bit active.

hoglet67:
> Could you have a look at updating the CPLD to add the things discussed above:
> 
> 1. Remove 2 msbs of delay
> 2. Make psync run during hsync
> 3. Make Vsync appear on psync when version bit active.

I've just pushed a CPLD that should do that.

The psync changes are untested, so let me know if they don't work as expected.

IanSB:
> The psync changes are untested, so let me know if they don't work as expected.

It seems to work in that I can now capture a whole line when the last pixels are after the start of sync but it seems to have introduced a problem:
When I adjust the delay setting on the ZX81 in positions 0 & 1 I get good results but in positions 2 & 3 I get psync jitter.
Is psync now running all the time as that would explain it because there is going to be some potential jitter when going from the trailing edge of Hsync to waiting for psync.
Previously psync stopped on the leading edge of hsync and didn't restart until some time after the trailing edge. I think what's needed is for psync to continue during hsync (or at least a few cycles after hsync starts and then stop, resuming some time after hsync as before.
Also as Vsync is now available, we could drop the xnor of hsync and vsync as that leads to some double negatives when checking polarity of the syncs and I can wait for Vsync directly from the switched psync source during blanking.

hoglet67:
> Is psync now running all the time as that would explain it because there is going to be some potential jitter when going from the trailing edge of Hsync to waiting for psync.

What's happening now is the rising edge of Hsync is resetting the counter to a small negative value (including delay). It then counts up. The value is currently 111 <d1> <d0> 100. Think of the MSB as a sign bit. So when delay is 3, for example, the value is -4.

While the counter is negative, psync is held in reset.

Once the counter reaches 0, the first Psync edge happens, which should always be 0->1.

What I suspect is happening is with larger delay values Psync starts too soon after Hsync, and is missed. (i.e. the firmware needs a certain amount of time after it sees Csync go high).

In fact, when you were previously programming delay values of 0..3 with a 4-bit delay, that's actually giving you ~12-15 pixels of delay (because the counter counts up).

Let me try to change it to work like this (which should then be identical to v5).

(I'll just make this one change for now, and leave the xnor change until later)

Dave

hoglet67:
That's now pushed.

IanSB:
> That's now pushed.

That has similar issues but I've looked at the signals on a scope and can see what's happening.
Yellow is sync, cyan is psync:

Delay0:
delay0

Delay1:
delay1

Delay2:
delay2

Delay3:
delay3

Delays 0-2 have jitter because there is a Psync transition immediately after sync goes high, Delay 3 is OK.
I think perhaps a software solution is going to be required for this, maybe a few hundred ns delay after sync goes high before looking at psync?
I suppose the other problem with this is making sure the delay on psync is adequate for a wide range of clock frequencies (in both software and hardware). Probably the lowest pixel clock would be 3.579Mhz (VIC 20?) or 4Mhz (Ohio superboard) although I'm currently testing with 6.5Mhz (ZX81) and the highest at the moment is 28.32 Mhz (VGA text) although I'm going to try 36 Mhz sometime!

hoglet67:
That helps...

I though the software waited for the first 0->1 psync transition after csync went high? Sounds like the case where psync is already high is being treated as an edge?

If the polarity of sync really doesn't matter, the there might be a hardware solution, which is when the counter is reset by the rising csync, then simply leave psync at whatever level it's at. I can still see some possibilities for race conditions.

IanSB:
> I though the software waited for the first 0->1 psync transition after csync went high? Sounds like the case where psync is already high is being treated as an edge

Looks like it assumes a low and acts on the next high it sees.
I could change it to look for a low first and that should fix the above but I think I'd have to test that across a wide range of clock frequencies and even then there might be race conditions where there is a spike after csync depending on the relative phase of the signals so maybe a small delay to avoid looking during that period would be more reliable.
I'll do some tests later today unless you have any other suggestions. (I'll also try the old v6.0 in case the shorter delay is adequate to avoid wasting a register)

IanSB:
I looked at this with the 28.3 Mhz pixel clock:
This is v6.0 firmware:
v6 0

This is v6.1 firmware:
v6 1

Rather than having a fixed delay I could look for a low on the psync line for a period determined by the current pixel clock frequency and any glitches would simply restart the clock on that timer.

hoglet67:
Are you happy that you understand what's happening here? I'm pretty sure I do now.

All of the 6.x CPLD versions force psync low while the counter is negative, so that the first psync edge after the delay is always a 0-1 edge. So the above waveforms are as expected, given the phase of psync.

It appears that the software is very quick to transition from WAIT_FOR_CSYNC_1 to WAIT_FOR_NEXT_PSYNC_EDGE. So fast that it just manages to catch this edge.

The only reliable solution I can thing of is to insert a short delay in the software.

Or go back to the previous scheme, where psync is held low for the whole of the hsync pulse.

IanSB:
I've been looking at the timing issue and optimising the code around the csync to psync transition and experimenting with different delays in the CPLD.
One of the possible issues is the time between the trailing edge of csync and the first psync as if that's too long some pixels may be missed but if it's too short there may not be enough time to switch from detecting the end of csync to detecting the first psync. This is particularly important with EGA mode where there is virtually no time for front or back porch blanking and as already mentioned the only card I have that produces EGA timing has the video going 1 char beyond the start of hsync. I think that is probably not standard timings so I would have to capture the timing at least 1 and probably 2 chars before that. With my current tweaked CPLD I can capture up to 3 chars earlier so hopefully that will be enough.
One thing I noticed is that the unprocessed version of csync is being fed to the Pi
i.e. csync <= csync1 if  it was csync <= csync2 which is delayed by 4 clock cycles that may move the glitches to the left of the csync rising edge.
Also if using the deglitched version of csync would it be sensible to drop the second read of csync state like we dropped the second read of psync state?

hoglet67:
I'd also previously noted that we might benefit from sending csync2 to the Pi. I'd then promptly forgotten about that. Feel free to give it a try.

I've got a few free days this week which I was going to spend on the 1MHz Bus FPGA adaptor. But I'll follow closely what you are doing, and feel free to prod me if there's anything you would like my help with.

Dave

IanSB:
OK I'll carry on experimenting.
Can you point me to some test software to try the fine h scroll as that's in the Csync to Psync transition and needs to be checked.

hoglet67:
I use Tricky's RallyX demo:
https://stardot.org.uk/forums/viewtopic.php?p=124091#p124091

Once loaded, you need to type U S B on the keyboard to correctly set the sync timing. Then you can scroll around with the cursor keys.

The horizontal scrolling should be super-smooth. If it's not, then something is amiss.

PR #63, opened 2019-03-29
Full file based profile system with autoswitching

IanSB:
This set of commits updates the profile system to work entirely from files with no hard coded settings.
Also updates the CPLD to V6.3 (removed 1 register to reduce the time from sync to psync and send csync2 to the pi)
The software should default to BBC settings if no profile folder found.
Please delete your existing profile folder and replace with new one.
Folder structure:
Individual profiles are stored as .txt files in the profiles folder. Switching sets are stored in sub-folders in the profiles folder
The individual profiles and the folder names of the switching sets appear under the Computer profile setting. If a switching set is selected then the contents of the switching set folder appear under the Mode Sub-Profile setting. If the current source matches a file in the switching set it will be auto selected.
Note there is a huge variation in timings/pixel clocks for nominally the same standard so there will have to be many PC switching sets, one for standard IBM timings and others for "compatible" - ish cards.
I have two cards and they both have different timings/pixel clocks for the same modes.


hoglet67:
Nice. I'm out in the garden all day, but will try to merge this evening.

hoglet67:
I've got another full day in the garden tomorrow (rebuilding the pergola) so probably not going to get around to trying this out until Sunday now.

I've got no changes of my own pending, so there's not going to be any issues merging.

PR #64, opened 2019-04-03
Support separate H and V syncs

IanSB:
Support separate H and V syncs and fix issue with cache preload taking up 1 line period by waiting for hsync


IanSB:
I've added variable scanline intensity from 0-15 0=black and 15=barely noticeable.
Note this uses the menu half of the palette to create duplicate entries for all colours with a different intensity so scanlines are disabled when the menu is on screen to avoid a messed up display.

PR #70, opened 2019-04-05
Resolution and interpolation adjustable from menus

IanSB:
Resolution and interpolation are now adjustable from the main menu and there is now a resolutions profile folder as well so you can specify as many resolution options as you like.
There are three interpolation options: Off, Medium and Soft which correspond to scaling kernel 8, 2 & 6 respectively. (1 & 4 have bad artifacts and there didn't seem to be much difference between 3, 5, 7 and 6.)
When one of the above options is changed, a warning is printed at the top of the screen that a reboot is required and when the menu is exited the config.txt file is rewritten and the Pi is rebooted.
Note that config.txt is re-created from "default config.txt" each time so any permanent config changes should be made to that file.
Also the profile is now stored by name in a profile.txt file instead of cmdline.txt so it can support spaces in filenames.
New files/folders required on Pi SD card:
Resolutions folder
Profile.txt
Default Config.txt  
Cmdline.txt is no longer used.


hoglet67:
This is great, and addresses one of my annoyances when switching between monitors, which I frequently do.

A couple of questions:

1. What happens if you select a resolution higher that the monitor can support? Does the Pi firmware blob do any sanity checking here against the EDID info (I assume not).

2. What's your feeling now about reading the EDID information? Is that still worth doing?

3. I have a strong dislike of filenames containing spaces, coming from a Linux background where these are generally discouraged. I think we should replace all spaces with underscores, and then simply map underscore to space when displaying the filename. Do you have any problems with that?

4.  Did you convince yourself that resolution is orthogonal to profile, rather simply than being another part of a profile?

5.  I'm wondering whether every profile should be a folder. It's more uniform, and it would allow things like resolution and saved settings to be held together. Screen shots would also be kept in the same place, specific to the system, which would make the card easier to manage. Thoughts?

IanSB:
At the moment if you select a higher res than the monitor supports there will probably be no output but not tested. Perhaps "Auto" should be the default from the resolutions list (Auto contains no hdmi setting and this seems to make the Pi pick the resolution using EDID). 
Also I think I'll add an option that if two buttons are held down during bootup it will revert to Auto to recover from a bad mode setting.
There are an awful lot of HDMI registers although I don't know how many of them change for a mode change:
https://github.com/christinaa/rpi-open-firmware/blob/master/broadcom/bcm2708_chip/register_map_macros.h
The space / underline switch might cause issues with user created folders and files that contain a space.
If every profile was a folder then I couldn't determine the difference between switching sets and standard profiles. I think mixing user generated screen caps in with profiles is a bad idea as those will likely be deleted / move etc and that might lead to accidental deletion of profiles. Having a screencaps folder in the root with sub-folders named for each profile might be better.

hoglet67:
> There are an awful lot of HDMI registers although I don't know how many of them change for a mode change:
> https://github.com/christinaa/rpi-open-firmware/blob/master/broadcom/bcm2708_chip/register_map_macros.h

What I did to discover the other registers was simply to dump them all on boot, and look for values that changed only when the mode is different.

I don't think this is particularly important though, as the reboot is very quick. The only downside is that currently the calibration is lost. But that's will be sorted when we can persist config changes.

> The space / underline switch might cause issues with user created folders and files that contain a space.

The only problem case I can see is when a user has both:
- BBC Micro
- BBC_Micro 

Is that the specific case you were thinking of? It seems a bit unlikely.

I'm not trying to prohibit spaces. I would just rather we didn't have to include them in files we include as examples.

> If every profile was a folder then I couldn't determine the difference between switching sets and standard profiles

Isn't this just down to the number of profiles that are discovered in the folder?

If there are multiple profiles, then you have a switching set?

Maybe I'm missing something here about the difference between a top level profile and a switching set?

IanSB:
In order for the underline/space to work properly while still allowing users to use spaces, the profiles and sub-profiles will need to maintain two arrays of strings, one with filenames and one for display with underscore converted to space. At the moment there is only one array which is used for both. Not a problem but will have to be done as you can't convert back and forth while still allowing users to use spaces.
I don't really see the point of using folders for all profiles unless we actually put something else in there as it's rather redundant to use /profiles/computer/computer.txt.  If you look in the profiles folder at the moment you can immediately see which ones are switching sets. Also from a practical point of view it makes maintaining the files easier as there are less levels to traverse.
As mentioned I don't think it's a good idea to put screencaps in there and even if we make the output resolution adjustable on a per profile basis I don't think it's a good idea to put resolution files in each profile folder as that's just adding duplication and also increasing maintenance hassles. (e.g. you shouldn't just put 50Hz modes in the BBC folder as someone might need to use it on a 60Hz only display)

hoglet67:
I do get your point about folders, if there is only one thing in there. I had imagined the screencaps going into a sub-folder within the profile, but a separate hierarchy probably makes sense. It would be nice to keep screencaps from different profiles in separate folders I think, but that's entirely doable.

Re: the underscore/space thing, I was imagining just one array with the actual filename, and a quick underscore->space replacement when it was displayed. Is there ever a need to go from the display name back to the file name? I really do hate being forced to put spaces in filename. That's 30+ years of using Unix systems for you. ;-)

Issue #69, opened 2019-04-04
Feature: Save changes back to the SD Card


hoglet67:
Changes are possible under 4 menus:
- processing
- settings
- geometry
- sampling

The challenge is:
- tracking what has changes
- deciding which level of profile a particular change belongs in

By level of profile, we mean either:
- /Default.txt
- /SomeProfile.txt
or
- /Default.txt
- /SomeProfile/Default.txt
- /SomeProfile/SomeSubProfile.txt

A possible starting point is that all changes should be saved to the most specific level of profile that currently exists. i.e. the default files are manually maintained.

Can we think of any exceptions where this is clearly wrong?

IanSB:
I've been thinking about and there are issues with all the different approaches.
One of the primary reasons for the default folder is to make maintenance of the default profile set easier as in many cases only the default will need to be updated and I agree that /Default.txt should be manually maintained with any changes to a standard profile saved to /SomeProfile.txt

The problem comes with the switching sets: 
In all cases, geometry and sampling changes should be saved to /SomeProfile/SomeSubProfile.txt
The problem happens with where to save other settings. In order to allow temporary changes to another setting like scanlines,  /Default.txt and then /SomeProfile/Default.txt are only read once when the switching profile is loaded and as each  /SomeProfile/SomeSubProfile.txt is auto selected (which normally have just geometry and sampling) it allows the other settings to persist across the auto-select.

If we save those other settings to /SomeProfile/SomeSubProfile.txt then each time a sub-profile with those settings is auto-selected you will lose the persistence of the temporary change and in the case of the auto-select on/off setting that would make that option virtually unusable.

To work around this, the other (non-geometry/sampling) options should be saved to /SomeProfile/Default.txt so they are common to all sub-profiles.

There are some circumstances where it would be preferable for one sub-profile to have a different setting to the other sub-profile and I think for the moment that will have to be managed manually especially because if you make a custom change to one sub-profile you have to put the opposite setting in all the other sub-profiles to make sure the setting is undone when they are auto selected.
(This already is being used to switch genlock off for the 70Hz "CGA/EGA on VGA"  modes) 

Are you going to look at the saving code or do you want me to do it?

hoglet67:
> Are you going to look at the saving code or do you want me to do it?

It's about third on my list things to work on at the moment, so if you want to have a go then feel free.

IanSB:
I think this can be closed now as outstanding issues are covered by 
[...]

PR #71, opened 2019-04-07
Profile saving, power on mode reset, vlock adjust, status info & remove spaces. Addresses https://github.com/hoglet67/RGBtoHDMI/issues/69

IanSB:
These commits implement the following:
Underline used in filenames and converted to space for display.
Vlock adjust option added which controls the range of the genlock, optionally restricting to +-5% to avoid creating wildly out of spec timings which might not be visible on some monitors.
If any two buttons are held down during reset, the display mode is set back to Auto and interpolation is set back to Off.
Status info is now shown in the blank line under the menu title and includes warning about lock failure if input and output frame rates mismatch.
Also helper info showing current clock, PPM error, lines per frame, sync polarity and source freq are shown when geometry settings of clock freq, line length, clock tolerance, lines per frame and sync type are adjusted.
Profiles are now saved using the Save Configuration option on the main menu.
Any changes to a standard profile are saved to /SomeProfile.txt
Any changes to a sub-profile are saved to /SomeProfile/SomeSubProfile.txt
Except for the auto switching setting which is saved to /SomeProfile/Default.txt to allow autoswitching to be enabled & disabled across all sub-profiles. As this setting is handled differently it has been moved to the main menu and placed with the profile settings themselves.
The loading and saving system has been rewritten and now uses the strings from the menu system converted to lower case with spaces replaced by underlines (e.g. the Num Buffers setting is saved as num_buffers=x  etc.
The geometry and sample tables are always written, but the other settings are only written to the file if they differ from the default.txt
Replace all support files with ones from these commits:
config.txt
default_config.txt
profile.txt
Profiles
Resolutions


hoglet67:
I've given this a quick try, just with the BBC, and it looks very nice.

I'll do some more extensive testing later.

hoglet67:
I've been doing some more testing this morning, and it's all looking very nice.

I've just pushed a couple of small commits: 99c998ae and cb420b56.
 
In the second of these, I've switched the 12x20 font to the SAA5050 US ASCII version (the SA5055), because the _ character (95) was a # in the UK version, which was confusing when a profile was saved.

I've also found one crash, when trying to use 1280x720 mode (and other similar ones). There's a separate issue for this #72

I've also added #73 #74 and #75 onto the TODO stack.

A couple of other things came up during testing, mostly not related to this saving feature, but I'll jot them down here anyway.

1. The return and keymap properties are not configurable through the UK. These are the only examples of such properties, so we should probably fix this. (#76)

2. The default Auto resolution ends up choosing 1920x1080@60 Hz on my Pi. I guess that's down to the EDID data in the monitor. So probably not much we can do. I might be worth understanding how the Pi decides what mode to actually use in this case.

Again, thanks for doing this work, you've done a very nice job of it!

PR #83, opened 2019-04-09
Add sync timeout & fix scaling + OSD bugs

IanSB:
There is now a sync timeout on both H and V sync which allows the ZX80 to work properly with the screen blanking when no sync is generated. Also it fixes a problem on the PC where the Pi hung up when the Pi and PC were powered at the same time. (The Pi detected the wrong sort of V sync at startup and then hung waiting for pulses that never arrived)
There are also various bug fixes.
BTW I sent out a few boards to PC testers so maybe we should think about getting a release ready and updating the documentation.

### DIFF
--- a/src/defs.h
+++ b/src/defs.h
@@ -232,6 +232,9 @@ typedef struct {
 #define MAX_RESOLUTION 64
 #define MAX_RESOLUTION_WIDTH 32
 
+#define FRAME_TIMEOUT 24*1024*1024           // ~24ms which is over a frame / field @ 50Hz (20ms)
+#define LINE_TIMEOUT 74*1024                 // ~74uS
+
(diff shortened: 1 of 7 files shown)


hoglet67:
I've just merged this.

I was wondering about an option to choose between a black screen (as you currently have) and some kind of a "no signal" banner screen. On a ZX80 you clearly need the black screen, but on other systems it might be helpful to know that the system is still powered up, but for some reason isn't putting out valid video.

IanSB:
The psyncs are still running so the capture area is written with whatever the RGB inputs float at. On my system that varies from a black to a coloured screen. With the sync timeout, the menus now work with no signal so I could put a message about no signal on the status line.

PR #85, opened 2019-04-10
Profile system updated with restore default, screen caps folders & sync loss message

IanSB:
The profile system has been updated to write saved profiles to the "/saved_profiles" folder instead of overwriting "/profiles". If the system finds a profile in saved_profiles with the same name as in profiles it is used in preference.

The restore default option simply deletes the copy in saved_profiles which allows the old one in profiles to be used instead. Also if you try and save a config that is identical to the default it actually deletes any saved_profile of the same name. This means that saved_profiles should only have actual custom configs.
To make those configs permanant, manually move those files into profiles overwriting the originals.

Note the comparison of saved profiles with default is a little fragile as it is simply comparing the text files so a manually edited file with a different order but the same actual settings will appear to be different.
To avoid problems, only auto saved files should be used in release profiles.

Also to avoid issues with comparison, the Invert setting has been removed from the sampling menu as it was never a user setting (it is still shown in the info menu for debugging purposes)

Other changes:
Screen caps now saved in /Captures/profilename
Loss of sync detected and a warning message shown on the status line

Please update your profiles to the ones in this commit


hoglet67:
Just a quick question....

If the presence of a Saved Profile file means the Released Profile file is never actually read, doesn't that make it difficult for us to make any changes to the configuration files from release to release?

I guess we would need to state whether a particular release requires users to delete their saved profiles.

IanSB:
The code could be tweaked to process the release profile before processing the saved profile instead of substituting it, that way any new settings would be picked up. (This is an extension of the existing processing  of the default then the selected profile)

hoglet67:
In terms of changes to the configuration system over time, I think there are four cases to consider:
a) A property is added - your suggestion covers this case 
b) A property is deleted -  it's harmless if the property still exists in a saved profile
c) The format of a property changed
d) The value of a property is changed

It's (c) and (d) that I'd like to think about some more.

With (c) and (d), because all properties are currently included in the saved profile, the updated value of the property in the new release profile would be ignored, even if the user had not modified the property from it's default value.

This is actually quite limiting - it prevents us adding palettes (unless at the end of the list), or adding new choices to an existing option, or reordering choices.

It would be nice to have a utility class that acted a a small property database. It just needs to store name/value pairs, and have the ability to lookup a property by name, and the ability to modify the property value.

When generating the saved profile from the current profile, it would be possible to compare each property with the value in the release profile database, and only actually include it if the value was different.

There's probably more to it than that, but it doesn't seem that hard.

IanSB:
> With (c) and (d), because all properties are currently included in the saved profile, the updated value of the property in the new release profile would be ignored, even if the user had not modified the property from it's default value.

I think that is going to be a problem whatever approach is used. In the prior version, if you saved by overwriting the existing profiles with new ones any customisations would be lost entirely when you upgraded but using the saved profiles option may allow the customisations to be propagated.
I think the best way to ensure reliable updates would be to include a version=x entry in the profile so the software could update any customised profile to the latest format if that had changed significantly like changing palette from =n to =string.txt

The reason I prefer using a substitute profile rather than another heirachical profile is that the saved profile can be made the new default by simply moving it to the profiles folder. I view the saving system as something to use for experimentation which is why it is easy to save and revert. When you are happy you move the result to profiles. Might be worth adding another menu option to do just that (with some additional confirmation to make sure you really mean it).

hoglet67:
Merged work so far.

I've just been reviewing save_profile and spotted a few things, some minor/cosmetic, some less so.

First, the more significant things.

**1. It seems the configuration property name is now mechanically derived from the human readable name.**
```
            strcpy(param_string, features[i].name);
            for(int j = 0; j< strlen(param_string); j++) {
                param_string[j] = tolower(param_string[j]);
                if (param_string[j] == ' ') {
                    param_string[j] = '_';
                }
            }
```
This is really not a good thing to do. It means that cosmetic changes to  how properties appear in the UI now also affect the configuration profiles. It also means that if someone wants to make a German version of the UI, all the property names change. It also means you end up with horrible names like:
```
input_mux_(3_bit)=0
```
I think we really need to add a field to the param_t struct, so that the two values are distinct:
(a) property name (used by the configuration system)
(b) human readable label (used by the UI)

I would call (a) "name" and (b) label.

**2. Comparing properties**

>   Note the comparison of saved profiles with default is a little fragile as it is simply comparing the text files so a manually edited file with a different order but the same actual settings will appear to be different.

I see that you are already using strstr to only emit properties that differ from the Default profile. This goes some to alleviating my earlier concern, as properties that are only defined in the Default profile can be changed between releases.

I don't see how the current code is sensitive to the ordering of properties? Can you give an example?

> To avoid problems, only auto saved files should be used in release profiles.

If this is true, I would like to find a work around. Mostly because I would like to be able to include comments.

One easy way to make the system more robust is to generate a "reference" profile that is used for comparison. This could be constructed as follows:
a) load the Default Profile
b) load the Selected Profile (optional)
c) load the Selected Sub Profile (optional)
d) call save_profile with an option to write all properties to a reference_buffer
e)  load the Saved Profile

When Save Configuration is now called, the reference_buffer could be used in the strstr comparison

I think this has two major advantages:
- It's immune to cosmetic differences (like white space, and trailing comments)
- It's easier to control exactly what differences are included in the saved profile. If you move (b) and (c) to after (d), then I think it works exactly was your current system does (only a bit more robust). If you do (b) and (c) before (d), then the Saved profile contains only the things the user has actually changed.

I do understand your point about wanting to using the Save Config system as a profile authoring tool, hence wanting the saved profile to be the complete set of deltas from the Default Profile.

My preference at the moment would still be that is is the deltas from the Default + Profile + Sub Profile (i.e. only real differences). I prefer this, because it reduces the potential for problems as things change from one release to the next.

But I'm willing to try the system as it currently is for a bit.

**3. A few  minor/cosmetic suggestions save_profile make it more readable, and to reduce the number of lines of code.**

a. Replace:
```
    for(;;) {
        if (features[i].key < 0) {
          break;
        }
```
with
```
    while (features[i].key >= 0)  {
```
b. Replace several occurrences of this:
```
            sprintf(pointer, "sampling7=");
            pointer += strlen(pointer);
```
with:
```
            pointer += sprintf(pointer, "sampling7=");
```
as sprintf returns the number of characters written.

c. You could use a loop to emit the mode7 variants of sampling7 and geometry7.

PR #86, opened 2019-04-11
Fix some issues in review

IanSB:
Contains the following changes:
Add property_name to param_t.
(Most of the names remain as before but I tidied up a few so replace your profiles.)
Requested readability improvements in save_profile
File save code is now property order independent when comparing with default.

> You could use a loop to emit the mode7 variants of sampling7 and geometry7.

I did it that way because the ultimate intention is to replace mode 7 with profiles and a separate block makes it easier to remove.


hoglet67:
Thanks for taking the time to do this Ian, much appreciated.

BTW, I've been working on the a new version of the Atom RGBtoHDMI PCB this week. It's the same circuit as before, but I've re-laid out the PCB to fit snugly  in a Evatron EVA73G case. Remembering out conversation last week, I've stuck to DIP components throughout, apart from the CPLD and three decoupling capacitors. I'll post some images on Stardot tomorrow, and send the PCB off to JLCPCB over the weekend.

Issue #87, opened 2019-04-13
Review contents of processing and settings menus


IanSB:
I think we need to determine what each of these two menus is for and review what items belong in each one.
I think the processing menu should contain options that affect the current profile and that the user might want to change to customise their preferences and the settings menu should contain elements that are generally only changed when switching profiles or rarely changed for testing etc.
e.g should palette move to settings because it only changes for specific profiles while palette control remains in processing?
Similarly should scaling move to processing as it is a user preference
Should the menus be renamed?

hoglet67:
The split I had originally in mind was was:
- Processing - configuration related to additional processing of the actual video content in some way
- Settings - configuration related to how the system operates generally

So with this definition, scaling probably should go into the processing menu.

By all means lets talk about changing this though.

But it has to make sense as far as the user is concerned. From your description above I'm struggling with that aspect of it. We should be making this sort of change to improve the user experience, making it easier for them to find a feature they are looking for,

Why not sketch out what would they would be renamed to, and what would be in each?

IanSB:
I was thinking along the lines of:
Preferences: Things a user would want to adjust when within a profile)
Settings: Things that have to be set to make the profile work and if you change them they will mess up the profile:
So:
Preferences:
Palette control: (enabling/disabling in band control and NTSC artifacting)
Scanlines:
Scanline Level:
Scaling:
Mode 7 deinterlace

Settings:
Palette: If you change this you get wrong colours
Border Colour: (new feature I just added) primarily for zx80 white background
Vsync Type: changing gets jumpy display (electron)
Vsync indicator: only used for debugging really
Vlock mode: should always be locked really (if in & out freqs don't match it is inhibited anyway)
Vlock line: should be left at 10, adjusting only makes lag worse
Vlock Adjust: Changing makes wildly out of spect timings
Num Buffers: really should be left at single now that OSD and vlock issues sorted.
Debug: again only needed for testing

You might want to change the palette to get a monochrome, green or amber screen effect but that would require different palette entries for each of RGB, RGBI, RrGgBb, Atom etc so I think a better option there would be a new colour option in the preferences menu to allow the current palette to be converted to monochrome / green etc programmatically and that would also work with user palette files.

That leaves Input Mux which really feels like it should be in the Sampling menu as it is CPLD specific and associated with the 3/6 bit switch. It certainly would be in that menu if it was controlled by an internal bit rather than a GPIO.

[...]

PR #88, opened 2019-04-14
Add border, colour & invert options and fix issue with border scanlines

IanSB:
These commits add the following menu options:
Output colour: Normal / Monochrome / Green / Amber
Converts RGB to mono, or green/amber screen emulation (might need some work on getting green/amber colours exactly right).  This could be expanded to cover some of the other palette options like red only, not red etc as they currently ony relate to the standard RGB palette.
Output invert: off / on
Inverts the output
Both the above should work with any palette
Border: 0-127
Sets the logical border colour (the area that's never written to by the capture code). Mainly used for zx80/81 white background

Also scanlines in the border area are now handled properly. To see the difference, run the hdmi/nula demo before and after (especially the first 2 pics & the car).
Again update profiles to the latest

There is also a warning ("Interpolation recommended") put on the status line if any scaling option other than integer scaling is selected with interpolation off as that can look poor quality without interpolation.

Note: I moved the palette option into settings and scaling into processing to illustrate my suggestions in https://github.com/hoglet67/RGBtoHDMI/issues/87 but they can easily be moved back again.


hoglet67:
I'm seeing some issues with this on the Beeb (using the default BBC_Micro profile).

Initially the menu come up normally.

Pressing auto configure cause the border and whole screen to flash white (with scanlines) during the operation. The autoconfigure does work though.

After that's been done, the same artefact appears when ever the menu is brought up.

Happens in both Modes 0-6 and Mode 7.

IanSB:
That should be fixed now.
The menu half of the palette is now being used to display variable intensity scanlines which means when scanlines are enabled every other line has the top bit of each pixel set. Before displaying OSD info, scanlines have to be disabled and the top bit on those lines must be cleared by calling clear_menu_bits() so I have added that to the auto calibration and HDMI calibration code. Also the NO_SCANLINES bit wasn't being set when calling rgb_to_fb() from the calibration routines and it is now.

IanSB:
Just found an issue with the sync inversion bit while testing loss of sync on the ZX80 so don't merge just yet

IanSB:
I think that should be OK to merge now unless you find any other issues.

IanSB:
I've implemented the double width frame buffers and it mostly works but there is an issue with the frame buffer init code.
The FB size option has 4 settings:
Normal
Double Height
Double Width
Double Height+Width
All seem to work OK except Double Width.

When that is selected  init_framebuffer(capinfo) returns with the pitch set wrong with resulting messed up screen (it seems to follow the pitch of the previous successful call.
Here are the logs snippets for each option (using bbc mode0 on a 1600x1200 monitor 8bpp)
You must change the settings in this order:

Normal
INFO: Initialised Framebuffer
INFO: Size: 800x300
INFO: Pitch: 800 bytes

Double Height
INFO: Initialised Framebuffer
INFO: Size: 800x600
INFO: Pitch: 800 bytes

Double Width
INFO: Initialised Framebuffer
INFO: Size: 1600x300
INFO: Pitch: 800 bytes      <-- should be 1600

Double Height+Width
INFO: Initialised Framebuffer
INFO: Size: 1600x600
INFO: Pitch: 1600 bytes

if you reverse order and go back to Double Width you get the right value and a good image
INFO: Initialised Framebuffer
INFO: Size: 1600x300
INFO: Pitch: 1600 bytes

but the going back to Double Height crashes
INFO: Initialised Framebuffer
INFO: Size: 800x600
INFO: Pitch: 1600 bytes    <-- should be 800

If you save the config with the value set to Double Width and reset, it starts up OK
If Double Width is omitted from the options, the other three seem to cycle OK.

IanSB:
As it seemed sensitive to the previous size of the frame buffer I added a call to the mailbox to setup a dummy 64x64 frame buffer before calling the mailbox again to setup the actual frame buffer size and that seems to work. Any idea what's going on here?

hoglet67:
> As it seemed sensitive to the previous size of the frame buffer I added a call to the mailbox to setup a dummy 64x64 frame buffer before calling the mailbox again to setup the actual frame buffer size and that seems to work. Any idea what's going on here?

Absolutely no idea.

Possibly an obscure Pi Firmware bug?

Work-around seems reasonable.

PR #89, opened 2019-04-16
Crash bug fix

IanSB:
There was an error in the readability changes which results in a crash when the Oric profile is selected. Also some other minor changes


hoglet67:
All merged now.

BTW, could I get your view on #90?

PR #93, opened 2019-04-20
Genlock improvements

IanSB:
The existing genlock has issues with some monitors so I've been doing a lot of work to get reliable genlocking without the monitor losing lock and blanking the screen:
I've implemented acceleration and deceleration when changing the PPM error so there are no large jumps  and there are now 2 modes of operation:

Fast (2000PPM) mode which locks in similar time to the old code
Slow (1000PPM) mode which has more conservative timings and works reliably with some of my problem monitors.
The vertical offset period now measures the VSYNC position in addition to the active video which means the vlock line is now a larger value as it includes the vertical offset.
In order for the deceleration to work, there has to be at least 35 lines either side of the lock line so the default line is now 35 although that includes 21 lines of blanking so it's equivalent to line 14 with the old code.
Fast mode increases the lock speed 333PPM at a time until it reaches 2000PPM with a delay of 12 frames between each change. When it gets within 35 lines of the lock it starts decelerating 333PPM at a time in a similar way until the lock line is reached. Once genlocked, there is a window of +- 10 lines of drift allowed before a resync is performed and the resync is done at 666PPM and 333PPM.

Slow mode increases the lock speed 333PPM at a time until it reaches 1000PPM with a delay of 24 frames between each change. When it gets within 16 lines of the lock it starts decelerating 333PPM at a time in a similar way until the lock line is reached. Once genlocked, there is a window of +- 5 lines of drift allowed before a resync is performed and the resync is done only at 333PPM.

This latter mode works reliably on my problem monitor which previously would always blank the screen when a 1000PPM jump was made and even 500PPM jumps were not 100% reliable.

Note as the PPM error control is now done algorithmically, the discrete SLOWxxxx and FASTxxxx options in the vlock mode menu option have been removed.

There is also a bug fix for double width capture loops which messed up the debug and vsync lines.


hoglet67:
This looks like a good improvement.

I do have one small concern.

The +/- 1000ppm and 2000ppm options actually predated the genlock. They were put there for demonstration purposes, so you see what the judder associated with dropping or duplicating a frame in a particular game actually looked like. I actually used them quite frequently. Relying on unlocked for this is problematic, as the rate of drops may be slow and unpredictable.
 
Would be be a lot of work to put these back?

IanSB:
It should be straightforward to add them back in.
What order do you want the options to be in?
e.g. the original
   "Unlocked",
   "2000ppm Slow",
   "1000ppm Slow",
   "Locked (Exact)"
   "1000ppm Fast",
   "2000ppm Fast"

or something like:
   "Unlocked",
   "Locked (Exact)"
   "2000ppm Slow",
   "1000ppm Slow",
   "1000ppm Fast",
   "2000ppm Fast"

Where locked / unlocked are adjacent (this also makes the default.txt value for on/off 0 & 1 rather than 0 & 3

hoglet67:
I have a slight preference for the original order.

But I really don't mind either way.

IanSB:
I've added them back in retaining your order but I rotated them thus:

   "Genlocked (Exact)",
   "1000ppm Fast",
   "2000ppm Fast",
   "Unlocked",
   "2000ppm Slow",
   "1000ppm Slow"

There is no difference as far as the menu is concerned but it means the default entry in default.txt is 0

 I also renamed Locked to Genlocked and changed the following options from V Lock to Genlock because they only relate to the Genlock setting of V Lock Mode:
Genlock Line:
Genlock Speed:
Genlock Adjust

hoglet67:
Thanks Ian, that looks great. I'll merge later on when I have a moment.

Issue #68, opened 2019-04-04
Review use of BITCPLD_V1V2


hoglet67:
My concern was this is meaningless in the context of the Atom CPLD, and if the Atom CPLD ever reached V3 then there may be unexpected consequences.

It's not used much:
```
defs.h:#define BIT_OLD_CPLDV1V2      0x08000000  // bit 27, indicates old CPLD v1 or v2
macros.S:        tst    r3, #BIT_OLD_CPLDV1V2
macros.S:        tst    r3, #BIT_OLD_CPLDV1V2
macros.S:        tstne  r3, #BIT_OLD_CPLDV1V2
macros.S:        tst    r3, #BIT_OLD_CPLDV1V2
macros.S:        tst    r3, #BIT_OLD_CPLDV1V2
rgb_to_fb.S:        tst    r3, #BIT_OLD_CPLDV1V2
rgb_to_hdmi.c:        extra |= BIT_OLD_CPLDV1V2;
```

More generally, we need to be review anything in main code base that is conditional on the CPLD version. It's more than likely wrong to be doing it this way, because version numbers are highly CPLD specific.

Instead, we need to extend the cpld "driver" interface in cpld.h to provide the required information in a clean way.

The only place it's really safe to check the cpld cpld version is a cpld specific driver.

Lines to look at include:
```
geometry.c:97:   if (((version >> VERSION_MAJOR_BIT ) & 0x0F) <= 1) {
geometry.c:101:   } else if (((version >> VERSION_MAJOR_BIT ) & 0x0F) == 2) {
osd.c:1589:   if (((cpld->get_version() >> VERSION_MAJOR_BIT) & 0x0F) < 2) {
rgb_to_hdmi.c:722:   if (((cpld_version_id >> VERSION_MAJOR_BIT) & 0x0f) < 3) {
```

Issue #92, opened 2019-04-17
Improvement: Allow choice of OSD font (?) 


hoglet67:
It's quite visually jarring on the Beeb for the menu font size to change dramatically between Mode7 and Mode 0..6.

(I have a strong liking for the 12x20 sized font and would like to be able to use it in all modes)

At the moment, it's explicitly disallowed in 8bpp framebuffers:
```
   if ((capinfo->sizex2 & 1) && (bufferCharWidth >= LINELEN) && (capinfo->bpp < 8)) {       // if frame buffer is large enough and not 8bpp use SAA5050 font
```

I gave it a quick try (by getting rid of the bpp check) and it seems to work fine.

What are the failure cases that made you always use the 8x8 font with 8pp framebuffers?

IanSB:
It drops every other frame and also messes up the genlock a bit when the OSD is on screen.
Try running my modified ski slalom sideways scrolling test which is good for detecting frame drops. It also causes the red bar of the vsync indicator to glitch to the top of the screen. 

With the auto switching modes, a frame drop was seen as a timing/mode change which caused repeated glitching as it tried to determine the current mode although I implemented a workaround so that 2x frame timing was accepted as valid to stop most of that.

I think that the fast osd update will have to be reimplemented in assembler to get it working properly.
Have a look at the code generated by the compiler to see how efficient it is.

PR #94, opened 2019-04-21
Code tidyup

IanSB:
I've done a bit of code tidying up:
1. switched 3/6BPP and 4/8BPP table selection logic
Previously the sampling object selected the table based on the frame buffer bit size of 4 or 8bpp and the frame buffer capture code selected 3 or 6 bit loops based on the sampling size read out of the sampling object which was a bit backwards.
Now the sampling object selects 3 or 6 bit tables and the capture code selects the 4 or 8 bit loops.
(Please check with the Atom adapter)

2. Implement cpld->old_firmware_support to make the old firmware support independent of the CPLD type and version (addresses https://github.com/hoglet67/RGBtoHDMI/issues/68 )

3. Remove the odd and even bits from flags register R3 to free up some bit space. They were only used by the soon to be removed odd and even sampling modes so I modified them to have separate odd and even loops instead.


IanSB:
I've also added a font size option as mentioned in https://github.com/hoglet67/RGBtoHDMI/issues/92

The settings are:
Always 8x8 - always uses the 8x8 font
Auto 12x20 4bpp - will use 12x20 font if screen big enough and 4bpp mode
Auto 12x20 8bpp - will use 12x20 font if screen big enough in both 4bpp and 8bpp modes

NOTE: use in 8bpp modes causes every other frame to drop (see my sideways scrolling test) but doesn't seem to cause any other issues although I haven't done much testing
The default profile has been updated with new setting.

IanSB:
I think we should be looking to do a release soon as my PC testers have started to receive their boards and will be looking for software shortly. We need to do some regression testing to make sure it still behaves on the old CPLD and also the Wiki needs updating.

hoglet67:
Many thanks for doing the font option.

I agree now is a good time for a release. 

Looking at how I had named the previous releases, this one would be
- Early Adopter Release #6

Are you OK with that, or should we start to name releases?

The documentation does need a lot of work. Basically we need a User Manual (v3). Do you think the existing format is still appropriate:
https://github.com/hoglet67/RGBtoHDMI/wiki/RGBtoHDMI-User-Guide-(v2)
(possibly it should be split across several pages)

IanSB:
> Are you OK with that, or should we start to name releases?

Should be OK for now, maybe we should switch to names when the PCB is updated to properly support 6 bit mode. Also there are still a few things left to do like implement the proper in band palette control and full NTSC artifacting.
I've tested some simple NTSC artifacting code and it works but the quality is not great so I've been looking at the NTSC artifacting code in Dosbox as that does a good job but it may be too complex to do on the Pi.

PR #98, opened 2019-04-23
Update Atom palettes

IanSB:
Atom palettes updated based on recent experiments
There are 6 options:
   "Atom MKI Card",    black, dark green and dark orange all set to 0
   "Atom MKI Card Full", pedestal, dark green and dark orange enabled 
   "Atom MKII Card", 8 colour RGB TTL only
   "Atom MKII Card Plus", 8 colour + orange
   "Atom MKII Card Full", 8 colour + orange + dark green & orange
   "Atom 6847 Emulators", emulator palette 

Note the luma scale is now 207/256 rather than 81/100

I've called it the emulator palette because it seems to be used in more than just XROAR (Mess + Mame?) and seems to originate here: 
https://forums.bannister.org/ubbthreads.php?ubb=showflat&Number=64986

EDIT:
Both MKI options should give 4 level monochrome output
I also tested this by turning the colour saturation to 0 on my monitor and it looked correct.
Reverted back to 81/100 percentage scale as it runs faster


hoglet67:
Did you think about making ref_black a parameter to yuv2rgb(), then

"Atom MKI Card" could use 720
"Atom MKI Card Full", could use 770

At the very least, it would make it easier to experiment.

IanSB:
I put the parameter in but left it at 770 as 720 just desaturates the red & blue. You can experiment further if you want although make sure any new settings don't have any "off by 1" errors.

IanSB:
In several places when navigating up and down the menu heirachy the code called:
osd_clear(); (which briefly restored the palette to full strength)
redraw_menu(); (which dims the palette again)
This caused brief flashes on the screen if the two calls straddled a vertical sync.
I replaced those occurrences with osd_clear_no_palette(); to fix it.

hoglet67:
All merged now. Thanks for fixing the OSD flash issue. I noticed that for the first time yesterday (even though I'm sure it's been there for ages).

Issue #103, opened 2019-04-30
Sort Profiles and Resolutions in a consistent way


hoglet67:
At the moment, the menu order is the order the files were added to the SD Card. It gets very confusing when you add new profiles, or edit existing ones.

hoglet67:
Thanks very much Ian.

PR #105, opened 2019-05-05
Add 333ppm genlock speed

IanSB:
Add 333ppm genlock speed, (required for hdmi to vga converter)


IanSB:
I added alpha sort to profiles, sub profiles and resolutions which addresses https://github.com/hoglet67/RGBtoHDMI/issues/103

PR #106, opened 2019-05-15
Auto size the capture area depending on monitor resoloution

IanSB:
I've got the auto sized capture area working which tries to ensure that the capture window is always an integer multiple of the monitor resolution.

This means coloured borders on ZX81, PC CGA/EGA and ATOM will usually fill the screen with integer scaling without additional black borders at multiple resolutions.
To achieve this, the geometry settings have been changed to specify the minimum and maximum capture sizes instead of capture area and frame buffer size (which wasn't used in integer scaling modes anyway)
There are 3 new scaling options:
Integer Auto - picks a best fit capture area between the above min and max settings
Integer Min - uses the min setting as the capture area
Integer Max - uses the max setting as the capture area

I've updated the BBC, Atom CPLD and ZX81 profiles to use this new feature
Note to get the old style Atom CPLD setting on a 1080p monitor, select Integer Max

As an experiment I've set the minimum capture size on the BBC profiles to 640x256 and the maximum to 672x274. 
These settings will make no difference on a 1080p monitor which will select 672x270
On a 1600x1200 monitor 672x274 will be selected which ensures the whole line above and below the normal screen area will be captured so *tv1 and *tv255 are fully visible. Even more lines could be captured but that reduces the time available to draw the menu which means even the 8x8 font drops frames.

On a 1280x1024 monitor, 640x256 capture will be selected which will allow the screen to be filled although any mode except modes 0 and 7 will lose a few pixels on the right hand edge (at least until I can get the auto delay working). Of course interpolated options can be used in this mode to avoid that cropping if preferred.
 
Note that it is slightly more difficult to create a profile now as you have to think of a suitable range for min & max and how that will work at different resolutions but it means that once a profile is created there is less need for any tweaking to be done anyway.
Don't adjust the min and max settings with scaling set to auto as all the values interact with each other, select Integer min and set the min values and offset, then select Integer max and select the max values (offsets will be auto compensated for the max values)

Don't merge this just yet as there will probably be some further tweaks but let me know what you think.


IanSB:
I've added a new "Capture Area" option which can be Auto, Minimum or Maximum and removed the non-integer scaling option as it only really worked in 1600x1200 mode and even that was broken by the new capture area option. (It was nearly identical to the 4:3 option anyway)

The scaling options are now Integer, fill 4:3 and fill all

When scaling is set to integer, the capture area options work as before, i.e.:
Auto = pick best integer sub-multiple for capture area
Minimum = use minimum value for capture area
Maximum = use maximim value for capture area

When scaling is set to fill 4:3 or fill all the capture area works as follows:
Auto = use maximim value for capture area
Maximum = use maximim value for capture area
Minimum = use minimum value for capture area and mixmium value for frame buffer size.
This last option allows arbitrary scaling the same as was possible previiously.

I also fixed a fill 4:3 scaling bug with double size frame buffers as used by the Atom

IanSB:
I altered the Auto setting when scaling is set to Fill 4:3 or Fill All so that it uses a capture area half way between minimum and maximum. This gives additional choice by reducing borders and gives more pixels to actual screen data. You can still select maximum to get the full border.
I also changed capture area to capture size and updated the profiles again.

hoglet67:
> Don't merge this just yet as there will probably be some further tweaks but let me know what you think.

Ian, I've got visitors now until Sunday, so I doubt I will be able to try this until Sunday afternoon.

My main concern really is the added complexity, and how we eventually document this. Are there simple rules people should follow when setting the max capture size? Does the UI prevent the user from doing silly things, like setting min > max? Can min and max be set to the same areas?

How do the two capture areas relate to each other? I assume their the larger one is centred on the smaller one? Do the H/V Offset values apply to the min or the max area?

Is the BBC flashing cursor still ignored during auto calibration? I would expect this has got more complicated.

IanSB:
> Ian, I've got visitors now until Sunday, so I doubt I will be able to try this until Sunday afternoon.

No problem

> Are there simple rules people should follow when setting the max capture size?

You set the min size to the actual area of the active display or slightly larger if you always want a minimum border area and the max size can be anything equal to or bigger than that, up to the actual active area of the video system. On a beeb the minimum would be 640x256 and the max could be up to 800x288 which would be a full PAL image when using a 16 Mhz clock although I actually set it smaller than that to avoid capturing too many lines which made even the 8x8 font drop frames.
Generally speaking the min can't be smaller than the active pixel resolution and the max can't be larger than the transmitted signal. If you have a source without borders, then the max only needs to be slightly larger than the min, enough to capture the pixel shift in different modes and the *TV vertical shift. If you have a source with borders then the max should be set to capture the entire border which can be done visually.

The basic rules are:
Set Capture Size to Minimum, set min size to pixel area or slight larger to guarantee a border under all circumstances and adjust offsets and sampling delay for exact centre.
Set Capture Size to Maximum and adjust the max size until all the border is captured
Set Capture Size back to Auto

You can adjust the H & V offsets at any Capture size setting but minimum is best as it's easier to see when you are one or two pixels out.

> Does the UI prevent the user from doing silly things, like setting min > max?

Not explicitly as I found hard clipping could cause valid values in the one setting be lost if you messed around with the other setting. The values are silently clipped though so if min > max then max is set to min internally for the calculations.

> Can min and max be set to the same areas?

Yes, then it behaves just like the old system

> Are odd heights allowed?

No, the min and max heights are altered in steps of 2 so the vertical offset is always adjusted by a whole number

> I assume their the larger one is centred on the smaller one?

Yes

> Do the H/V Offset values apply to the min or the max area?

Both plus any calculated area. The actual value displayed relates to the min settings but the value is adjusted for the actual capture area internally so the image should always be centred after it has been set once at any capture size.

> Is the BBC flashing cursor still ignored during auto calibration? I would expect this has got more complicated.

The line skipping algorithm already takes into account capinfo->v_offset which is the adjusted value so it should just work. I tried calibration at 3 different resolutions and there were error free results in all 3.

I found it very easy to set up visually when updating all the profiles and in some ways it is now easier to centre the image exactly as you have to set a minimum capture area so it becomes obvious when you are out by 1 or 2 pixels.

PR #110, opened 2019-11-26
Six bit & analog PCB + CPLD v6.5 + software updates to support them

IanSB:
Hardware changes:
Six bit PCB v1 & v2
Analog interface v1
CPLD v6.5 with mono video on Vsync input using MUX setting
Software changes:
Updated profiles and palettes
Add new resolutions and replace Auto with Default@60Hz
Support for analog interface
Merge scaling and interpolation
Add warnings on resolution change and recovery option using menu-reset
Add cpld recovery option using 3 buttons + reset
Add calibration to front menu as well as button press (includes confirmation second button press)
Add confirmation second button press to restore defaults
Add screencap during menus using up+down buttons

All options a new user needs to setup are now on the main menu. i.e
Calibration, Resolution, Scaling, Frontend & Profile

Outstanding issues:
I added "Source Type" (Analog or TTL) to the sampling menu.
Initially this was going to be used to select the frontend but that was replaced with a global setting in the main menu. 
The option still has uses as it informs the profile which interface is required e.g:
Hide the DAC menu settings and setup the DACs with sensible defaults if TTL mode selected for that profile.
Completely hide non-matching profiles so that only TTL profiles are listed if no frontend fitted and only Analog profiles are listed if Analog front end fitted
Any thoughts on this?

The Pixel Sampling option in the geometry menu is obsolete and should be removed. It was left in as someone was using it with an old CPLD but now the programming option is available should we encourage users to update?


hoglet67:
Thanks for this Ian.

I'll merge tomorrow morning.

I'm definitely now of the view that we should encourage new users to update, with version 6.4 being the minimum supported version.

I'm also of the view that major releases (which this definitely is) should be installed on a fresh/blank SD card, i.e. we should not have to worry about backwards compatibility with old config files.

I'll answer your final two questions tomorrow when I've had a play (and re-familiarized myself with where things are).

hoglet67:
Hi Ian,

I've had a play with this, and merged your dev branch back into mine.

It's looking rather good!

A few small things I spotted testing (just on my Master):

- the border colour option breaks at 128 (and starts affecting the whole screen)

- when the return option is at the top, if you descend into a sub menu, then immediately press return, you have lost your position in the parent menu (i.e. you are back at the top again). This is different to when the return option is at the bottom. Is this intentional, or a bug?

- it would be nice to hide all the sampling menu options for the analog front end if this is not present (Frontend = None), as most existing users will not have this. This is quite easy to do if the optional are placed together at the end of the list.

- screen capture in mode 7 is a bit broken (maybe it aways was)

- screen capture in the menus (using left/right together) is a bit tricky to invoke. Is there a trick to doing this?

In terms of documentation, I has a couple of thoughts. I was thinking of the over structure being:
- A Quick Start Guide (as short as we can make it)
- A Reference Guide (probably organized menu by menu, explaining each option)
- A Tutorial for Creating a New Profile (for a new Machine) (maybe this is a stretch goal)

Dave

IanSB:
> the border colour option breaks at 128 (and starts affecting the whole screen)

This is deliberate and how the colour test profiles work.

> when the return option is at the top, if you descend into a sub menu, then immediately press return, you have lost your position in the parent menu (i.e. you are back at the top again). This is different to when the return option is at the bottom. Is this intentional, or a bug?

That was a deliberate change by me as I preferred it that way.

> it would be nice to hide all the sampling menu options for the analog front end if this is not present (Frontend = None), as most existing users will not have this. This is quite easy to do if the optional are placed together at the end of the list.

I'm thinking about doing that but it would have to be done carefully as the menu tables are also used to determine what is loaded and saved.

> screen capture in mode 7 is a bit broken (maybe it aways was)

Hadn't noticed that

> screen capture in the menus (using left/right together) is a bit tricky to invoke. Is there a trick to doing this?

It only works when the cursor is on the left of a menu, not highlighting an adjustable parameter.
Press and hold one button then press the other.

Issue #75, opened 2019-04-07
Update sampling config to use proper parameters


hoglet67:
Current sampling is a single parameter, explode this out to multiple parameters (with a common prefix like s_xxx samp_xxx or sampling_xxx).

Push the sampling parameter loading/saving down into the respective cpld implementation classes.

Issue #74, opened 2019-04-07
Improvement: Update geometry config to use proper parameters


hoglet67:
Current geometry is a single parameter, explode this out to multiple parameters (with a common prefix like g_xxx geom_xxx or geometry_xxx).

PR #111, opened 2019-11-30
Improvements to scaling / overscan / setup modes.

IanSB:
4:3 scaling has been improved and together with updated profiles should provide correct 4:3 aspect ratio scaling in Auto and Maximum modes. Minimum may not scale correctly as that depends on the x/y ratio of the minimum settings The halfway setting is affected as it uses the minimum setting.

Source size has been renamed to overscan and I also added "Setup mode" to geometry and sampling menus.
Previously, to setup certain values in geometry you had to change the source size (now overscan) setting which was messy and involved switching menus.
Now, there is a new option at the top of the geometry menu: Setup Mode with options Normal, Set Minimum, Set Maximum, Set Clock/Line.

When changed to Set Minimum, minimum overscan is forced and the h & v adjustments can be made.
Same for Set Maximum for Max h & v
When Set Clock/Line is selected, Clock Tolerance is forced to 0 internally so that all clock and line adjustments have an effect. Previously you had to manually force it to 0.
The Setup Mode setting is not saved and is forced back to Normal when you exit the menu.

There is a similar Setup mode in the sampling menu with options Normal and Set Delay.
Selecting Set Delay forces Minimum overscan so the delay can be set manually.
Again exiting  the menu forces setup mode back to normal.

I've also added some more info to the calibration summary which might help diagnosing problems with non-standard timings.

Also palettes updated.


hoglet67:
Hi Ian,

Just been looking at this pull request, particularly geometry.c.

I think there are there typos on the property names here:
```
+   {     H_WIDTH,        "Min H Width",        "min_min_h_width",       200,       1920, 8 },
+   {    V_HEIGHT,       "Min V Height",       "min_min_v_height",       150,       1200, 2 },
+   {    FB_WIDTH,        "Max H Width",        "max_min_h_width",       100,      1920, 8 },
+   {   FB_HEIGHT,       "Max V Height",       "max_min_v_height",       100,       1200, 2 },
```
I appreciate these are currently un-used, but I'm still thinking about whether we should try to implements issues #74 and #75. Positional parameters are a nightmare for us, and other users, going forward.

I'm finding the new geometry code hard to follow now, especially in: 
   geometry_get_fb_params(capture_info_t *capinfo)

Part of this is because I think these members of the geometery structure are now being used for the min/max capture area:
```
   int h_width;           // active horizontal width (in 8-bit characters)
   int v_height;          // active vertical height (in lines)
   int fb_width;          // framebuffer width in pixels
   int fb_height;         // framebuffer height (in pixels, before any doubling is applied)
```
I think it's worth renaming these, and updating the comments, otherwise there is a danger I'll completely loose the plot here.

Dave


  FB_WIDTH and FB_HEIGHT properties are sometimes the actual

IanSB:
Ok I'll fix that, I did some batch renaming to try to clarify things and it looks like it went wrong (like the min_min and max_min

IanSB:
That's fixed now.

hoglet67:
Thanks Ian.

BTW, I've just built up your v2 6-bit CPLD board.  It's working nicely, and it's nice to have free access to the SD Card. Bits for the analog board will be here on Wednesday.

Dave

PR #112, opened 2019-12-03
Atom/6847 support for analog six bit board

IanSB:
Support for multiple modes of operation with Atom/6847 on 6 bit board using modified analog board:

1. Atom CPLD updated to work on 6 bit board in conjunction with modified analog board: (pins changed plus support for inverted Y signal for direct connection to Y on 6847 for dragon/tandy etc.)
Note clamp signal is available on P19 which is routed to the analog board on pin 4 of the 4 way connector.
Also DO NOT PROGRAM THIS INTO ORIGINAL 3 BIT BOARDS as P19 and GPIO22 are connected together and are both outputs.

2. Support to run modified analog board with normal CPLD in both Atom and inverted video 6847 mode.
Profiles:
Atom_CPLD.txt -  Atom CPLD on Atom board or YUV6847 CPLD on 6 bit board
6847_CPLD.txt -  YUV6847 CPLD with direct (inverted) Y connection to 6847
Atom_Analog.txt - Normal CPLD with Atom Y video
6847_Analog.txt - Normal CPLD with direct (inverted) Y connection to 6847

None of the above are noise free on the 6 bit board so I suspect at least the clamp needs to be implemented. The direct 6847 connection seems to be even worse for noise so maybe some filtering needs to be done. 

There is already an anlog switch on the analog board so modifyin that to work as the clamp switch shouldn't be too difficult.


IanSB:
Messed up pull request

PR #113, opened 2019-12-03
Atom/6847 support for analog six bit board take 2

IanSB:
Support for multiple modes of operation with Atom/6847 on 6 bit board using modified analog board:

Atom CPLD updated to work on 6 bit board in conjunction with modified analog board: (pins changed plus support for inverted Y signal for direct connection to Y on 6847 for dragon/tandy etc.)
Note clamp signal is available on P19 which is routed to the analog board on pin 4 of the 4 way connector.
Also DO NOT PROGRAM THIS INTO ORIGINAL 3 BIT BOARDS as P19 and GPIO22 are connected together and are both outputs.

Support to run modified analog board with normal CPLD in both Atom and inverted video 6847 mode.
Profiles:
Atom_CPLD.txt - Atom CPLD on Atom board or YUV6847 CPLD on 6 bit board
6847_CPLD.txt - YUV6847 CPLD with direct (inverted) Y connection to 6847
Atom_Analog.txt - Normal CPLD with Atom Y video
6847_Analog.txt - Normal CPLD with direct (inverted) Y connection to 6847

None of the above are noise free on the 6 bit board so I suspect at least the clamp needs to be implemented. The direct 6847 connection seems to be even worse for noise so maybe some filtering needs to be done.

There is already an anlog switch on the analog board so modifying that to work as the clamp switch shouldn't be too difficult.


hoglet67:
Hi Ian,

I think the new "6-bit Atom/6847" CPLD should have a distinct (new) design number in the VHDL, which will allow it to have it's own unique "driver".  I don't really like the overloading of this into the original Atom CPLD "driver" (there is almost no commonality).

It also then means the software can then reliably this varient on power up.

Dave

PS. I have the analog board running now, tested with an Elk and a CPC-464. Working very nicely.

IanSB:
> I think the new "6-bit Atom/6847" CPLD should have a distinct (new) design number

That's done. 
Also as things were getting a bit confusing with the profiles for different CPLDs I split them into separate folders so the structure is now /Profiles/CPLD name/profiles

The selected profile for each CPLD is now stored separately so instead of "profiles.txt" in the root you now have profile_CPLD name.txt for each CPLD in the root. Each CPLD also has it's own default.txt

This means the original Atom build will only show the appropriate Atom profile and the new YUV6847 will only show Atom and 6847 profiles (so far - I'm going to look at the Spectrum 48K which is also YUV).
I might take this a bit further and hide profiles in each CPLD that aren't suitable for the current frontend.

> PS. I have the analog board running now, tested with an Elk and a CPC-464. Working very nicely.

That's great. Do you plan on building an Atom experimental variant?
I'm not sure of the best approach to a new PCB, whether to try and fit RGB and YUV capability onto one board or go for two separate boards. They are very similar but maybe not quite similar enough.

hoglet67:
I think that's much better, especially the seperating out the profiles.

I think one of the challenges we have now that we have so much flexilibity in the system, is keeping the user interface as simple as possible, by removing options that are not relevant to the current setup. Seperating out the profiles helps here.

I was wondering it if would also be worth giving the CPLD on your 6-bit board a seperate Design ID. Indeed, if the analog board was able to drive the "spare" connector pin, then the Design IDs 4 and 5 could be used, depending on whether the analog board was present. The wiring for this is already present on the V2 CPLD board. All that's needed is to  pull this pin high on the analog board.

This would then remove the need to manually select the analog board. It would also mean that we could have different profiles in play. That would be useful for the Amstrad CPC, which is usable without the analog board. It would also allow the analog board sampling options to only be shown when relevant.

(The only thing that caught me out yesterday with the analog board was finding the DAC options didn't have any effect, because I had the source type set to digital. My confusion was reinforced becuse the DAC settings and the Source Type are very far apart in the menu, so there is no clue that they are related).

> That's great. Do you plan on building an Atom experimental variant?

Probably not, as I already have the dedicated Atom adapter.

> I'm not sure of the best approach to a new PCB, whether to try and fit RGB and YUV capability onto > one board or go for two separate boards. They are very similar but maybe not quite similar enough.

I'm not sure either.

Sampling the Y signal on the Atom's PL4 with a fixed theshold is difficult for three reasons:
1. There is a lot of switching noise
2. There is a lot of crosstalk from other signals, especially if you have a long ribbon cable
3. The black level is not stable, and varies considerable between modes

It was only when I added a proper back-porch clamp that I got consistent results, regardless of the mode, the supply voltage, and how hot the Atom was.

There were a couple of questions I had for you:
- How well did the modded analog board + Normal CPLD work?
- Was the modded analog board + Atom CPLD noticable better, and in what way?
- Are they as good as the thought-hole Atom Adapter?

My belief is that to get the best result on the Atom's PL4 (and allow for a 1M cable) you need proper black-level clamp. It made a huge difference to the noise margins.

Possibly picking up Y directly from the 6847, may alleviate the need for a clamp, but it's a much lower amplitude signal, and that brings it's own problems.

Dave

IanSB:
> I was wondering it if would also be worth giving the CPLD on your 6-bit board a seperate Design ID. 

That might be useful, I think maybe rather than Normal CPLD it should be the RGB CPLD to contrast with the YUV CPLD.

> DAC options didn't have any effect, because I had the source type set to digital

My intention was to hide the DAC options when digital selected so that problem goes away.

>  if the analog board was able to drive the "spare" connector pin...

I'm now using P19 with the YUV CPLD as the clamp so driving it hard would cause problems when running that CPLD. There is provision for a pulldown resistor on P19 on the 6 bit board so if that was 10K then a 1K pullup on the analog board would override it and it would read high while still allowing the YUV CPLD to drive it without damage.

I assume P19 would directly affect the Design ID?

> This would then remove the need to manually select the analog board.

Still have to select which DAC to use (Which one did you use BTW?)

I've added the back porch clamp using the existing U5 termination analog switch and some other SMD components made into a 3D sculpture :)
>     How well did the modded analog board + Normal CPLD work?

Noise in the border when switching to games also noise when pressing reset 
> Was the modded analog board + Atom CPLD noticable better, and in what way?

Less noise (without clamp) than Normal CPLD
> Are they as good as the thought-hole Atom Adapter?

No, but once I added the clamp, performance was identical to the the thought-hole Atom Adapter
> Possibly picking up Y directly from the 6847, may alleviate the need for a clamp

Using the 6847 direct connection works but is noisier and even with the clamp it is still noisy.
I'm not too bothered by that as the standard Atom connector works and the option was intended for experimentation with Dragon 32 and Tandy CoCo which might not be as noisy as the Atom.

I'm also going to look at the Spectrum 48K as that has a YUV output bit I can't find any info on how many levels that uses.

PR #114, opened 2019-12-05
New 6 bit RGB CPLD with auto detect plus supporting software

IanSB:
New 6 bit RGB CPLD V7.0 with auto detect of analog interface and supporting software
Make sure R6 on Main CPLD board is changed to 10K then connect a 1K resistor from the P19 pin on the analog board to +3.3V or +5V.
The new CPLD is for the 6 bit board only and has ID 4 on it's own and an ID of 12 when the analog board is plugged in. There are new drivers for both IDs with irrelevant stuff removed like DAC settings in TTL mode.
The profiles have been updated with the new structure and there are now 5 folders:
Atom - through hole atom board
Normal - original 3 bit CPLD (do you want to change Normal to something else?)
RGB(TTL) - new 6 bit CPLD with no board
RGB(Analog) - new 6 bit CPLD with analog board plugged in
YUV - experimental YUV/6847 board with clamp on P19

Also only the right CPLD files are offered for upgrades and the software will think you are running a 3 bit board so to initially upgrade you will have to copy the v70 file from the 6bit folder to the 3bit folder in cpld firmware.
Note if the cpld is unprogrammed it defaults to the 6 bit folder.

I changed "frontend" to "interface" as it's behaviour has changed.
It will show a different setting for each CPLD so you know the capabilities:
Atom - Atom
Normal - TTL RGB (3 bit)
RGB(TTL) - TTL RGB (6 Bit)
RGB(Analog)  - Analog RGB/CVBS (UA1) or (UB1)
YUV - YUV (UA1) or (UB1)

In the first 3 modes, the setting can't be changed and is for info only
In the Analog and YUV modes you can switch between the UA1 and UB1 DAC types.
Note the UA1/UB1 selection is lost when changing CPLDs and may have to be set again when you reprogram the previous CPLD.


IanSB:
Note: Issue 1 six bit boards should be programmed with the 3 bit firmware to avoid confusion with analog board detection. Also avoids output/output clash with the YUV CPLD.

hoglet67:
Nice work Ian.

I've done the resistor mod, and the detection is working nicely.

A couple of thoughts about code maintainablilty....

1. For the VHDL designs, if we intend to keep the RGB_3bit and RGB_6bit designs in sync, we should use a common source vhdl file, with a generic parameter to control the small difference needed to change the version.

What you would end up with is:

vhdl/RGBtoHDM.vhdl
vhdl/RGBtoHDM.ucf

vhdl_RGB_3bit/RGBtoHDMI.xise
vhdl_RGB_3bit/RGBtoHDMI.jed
vhdl_RGB_3bit/RGB_CPLD_v65.xsvf
vhdl_RGB_3bit/fitting.notes
vhdl_RGB_3bit/iseconfig
vhdl_RGB_3bit/working

vhdl_RGB_6bit/RGBtoHDMI.xise
vhdl_RGB_6bit/RGBtoHDMI.jed
vhdl_RGB_6bit/RGB_CPLD_v70.xsvf
vhdl_RGB_6bit/fitting.notes
vhdl_RGB_6bit/iseconfig
vhdl_RGB_6bit/working

There is a way in the Xilinx project (.xise file) to set the value of a generic passed into the design. I use this quite a lot to build different configurations of a design from the same source files.

Unless you have plans for the RGB_6bit CPLD to diverge functionally, I think we should do this. I'm happy to do this if you agree.

2. We have now three seperate driver source files:

The files cpld_normal.c and cpld_rgb_ttl.c are 99% identical and could easily be the same. 

The differences between cpld_rgb_ttl.c and cpld_rgb_analog.c are also pretty miminal. i.e. there is far more code in common than there are differences.

It's possible to implement the three seperate drivers in a single common source file. This has three seperate parameter blocks at the end. And it needs one additional variable (supports_analog) to enable the additional properies when the analog interface is present.

I've just done this on a branch, so you can see what I'm talking about:
https://github.com/hoglet67/RGBtoHDMI/commit/3f97bb9ed599db87ab6753ff6281828ae9a8eda6

Again, as long as you have no plans to diverge the functionality of the RGB_6bit CPLD, I think we should go with this.

Finally, I agree the name "normal" is a bit wierd, and over time it will be incorrect. What about:

Currently we have:
```
cpld_atom.c:  .name = "Atom",
cpld_null.c:  .name = "Unprogrammed",
cpld_rgb.c:   .name = "Normal",
cpld_rgb.c:   .name = "RGB(TTL)",
cpld_rgb.c:   .name = "RGB(Analog)",
cpld_yuv.c:   .name = "YUV",
```

I can think of four options....

Option 1:
```
cpld_atom.c:  .name = "Atom",
cpld_null.c:  .name = "Unprogrammed",
cpld_rgb.c:   .name = "Original",
cpld_rgb.c:   .name = "RGB(TTL)",
cpld_rgb.c:   .name = "RGB(Analog)",
cpld_yuv.c:   .name = "YUV",
```
Option 2:
```
cpld_atom.c:  .name = "Atom",
cpld_null.c:  .name = "Unprogrammed",
cpld_rgb.c:   .name = "Beeb",
cpld_rgb.c:   .name = "RGB(TTL)",
cpld_rgb.c:   .name = "RGB(Analog)",
cpld_yuv.c:   .name = "YUV",
```
Option 3:
```
cpld_atom.c:  .name = "Atom",
cpld_null.c:  .name = "Unprogrammed",
cpld_rgb.c:   .name = "RGB_3Bit(TTL)",
cpld_rgb.c:   .name = "RGB_6Bit(TTL)",
cpld_rgb.c:   .name = "RGB_6Bit(Analog)",
cpld_yuv.c:   .name = "YUV",
```
Option 4:
```
cpld_atom.c:  .name = "Atom",
cpld_null.c:  .name = "Unprogrammed",
cpld_rgb.c:   .name = "RGB(TTL) v1",
cpld_rgb.c:   .name = "RGB(TTL) v2",
cpld_rgb.c:   .name = "RGB(Analog) v2",
cpld_yuv.c:   .name = "YUV",
```
I think I'm in favour of Beeb, because:
- it's simple
- a limited number exist (~20)
- almost all of those are used on Beeb (or Electron) systems
- if anyone wants one in the future, I would suggest they use your v2 PCB design, as the IDC connector is a much better option than the molex connector, and it is self-programming

Dave

hoglet67:
The same approach could be taken for cpld_atom and cpld_yuv, but there the case is weaker as:
- the original driver is much smaller anyway
- i don't know what the plan is going forward for the generic yuv interface, certainly the c and l filters were dealing with some very atom specific noise, and I'm unconvinced they actually have much of a benefit. So you might want to re-evaluate whether they are the best use of space in the YUB CPLD.

Oh, and there are sone nasty spaces in the directory name:  'vhdl_YUV_ 6 bit'

IanSB:
I think your option 2 but with the more formal 'BBC' rather than beeb.

I think a combined RGB / YUV interface may be possible but with some caveats to keep the component count down to fit on the existing board space:

1. The G (also Y) input would have to be permanently AC coupled which means clamping even in RGB mode. This could be achieved by sending inverted sync pulses out of P19 in RGB mode so the clamp would be active in part of the green blanking period. The clamp switch could be 1 part of a 4066.
(It's not possible to do back porch clamping on the generic RGB/Analog CPLD as the delay from the sync pulse to the back porch would vary depending on the pixel clock rate).
 
This would also mean a bit on the register chain to turn P19 from an input (so the analog board can be detected) to an output to drive the clamp.
Is that possible with the CPLD?

2. The RGB termination could be controlled by the other three switches of a 4066 (possibly without any extra resistors)

3. A 6 or 8 output DAC would be needed to control this: 2 for RB/UV, 2 for G/Y, 1 for sync, 1 for mono video on sync and 1 for termination on/off (could use the Data line to control that instead)  

BTW did you try sync tip clamping on the Atom instead of back porch clamping?
(i.e. did the back porch vary with respect to sync tip or did the whole video signal move up/down?)
 
I'm OK with you merging the VHDL, do you want to merge the drivers as well?

hoglet67:
> I think your option 2 but with the more formal 'BBC' rather than beeb.

OK, lets go with that.

> I'm OK with you merging the VHDL, do you want to merge the drivers as well?

Yes, that's just needs the branch to be merged.

I'll do that now, then do the VHDL.

Dave

hoglet67:
I've merged that now, changed normal to bbc and tidied the cpld header files.

Just pushed back up to github.

I'll do the VHDL now.

IanSB:
I think the version bit could be used to control the state of P19
When version is active, P19 is an input to read the interface
When version is inactive, P19 is an output sending inverted sync (clamp) pulses
Would that work?

hoglet67:
I've just pushed the changes that make vhdl_bbc and vhdl_RGB_6bit projects point to common source vhdl/ucf files. I've rebuilt the jedec file, and the fusemap part is identical in each case to what we had perviously. So this is really just a build change, rather than a functional change.

I don't think we should do the same with vhdl_atom and vhdl_YUV_6bit as I expect they will diverge over time (there are already some changed). So lets just keep them as seperate projects, and consider the atom version as frozen.

> Would that work?

Yes, you need to define clamp as an inout. Then you need to make the following changes:

Add a new internal signal
```
signal clamp_int;
```
Change your existing logic to use that:
```
          -- generate the clamp output
            if counter >= atom_clamp_start AND counter < atom_clamp_end then
                clamp_int <= '1';
            else
                clamp_int <= '0';
            end if;
```

And finally generate a tri-state output:
```
clamp <= 'Z' when version = '0' else clamp_int;
```

PR #116, opened 2019-12-10
Add pixel aspect ratio enforcement and fix Pi overscan "off by 1" bug

IanSB:
This update adds pixel aspect ratio enforcement and there are now 2 new options in the geometry menu:
H Pixel Aspect
V Pixel Aspect
The range is 0-8 for both
For a BBC or Amstrad H Pixel Aspect is set to 1 and V Pixel Aspect is set to 2 for 1:2 aspect
Atom is set to 1:1, Oric is set to 3:4 etc (these only affect Integer scaling)

The Integer scaler will always try to enforce the requested aspect although it may sometimes be impossible in which case it reverts to the previous approach of scaling x and y independently.
Note this means you will now get a correct aspect ratio "postage stamp" image when you select maximum or halfway in your previous example.
If you set at least one of the aspect values to 0 it reverts to the previous behaviour.

This will need extensive testing to make sure no new scaling bugs have been introduced.

One example to try is to look at the Amstrad CPC profile. This now has a minimum V of 200.
Previously if you set that it would scale 2:5 on a 1080p monitor (200x5 = 1000) and 2:6 on a 1600x1200 monitor (200x6 = 1200). Now it just scales 1:2 (aka 2:4 so 4x200 = 800). You can see the old result by setting one of the aspects to 0. This was originally worked around by setting the minimum V to 256 so it scaled properly but it was a bit of a bodge.
This should actually make the instructions a bit easier as there is no longer any need to think about compromises like the above 200 / 256 workaround.

For PC use, CGA is set to 2:5 and EGA to 2:3 

Note there is still a cosmetic menu issue when changing the overscan in "even" mode 7 but I think the other even mode 7 issues have been fixed.


IanSB:
The additional commits fix a couple of problems noticed when testing with an Oric 1:
The Oric 1 outputs NTSC timing for several seconds during bootup memory testing so the Pi gets the wrong timing for capture.  The Oric then changes to PAL output but that change wasn't being detected as there was no loss of sync and as a result the genlock wouldn't lock.
The windowing on H and V timing now works in all modes (previously only in PC autoswitch mode) but the window is based around the measured timing rather than the timing derived from the sub-profile so if the measured timing changes like NTSC to PAL it will cause a recalibration.

The vertical capture clipping also fixes an Oric problem:
The V offset is normally adjusted so that the minimum capture area is vertically centred on the active portion of the video. However if the active portion isn't vertically centred in the source video then the max capture area which is centred on the min capture area could extent off the end of active video and into the vsync area which would cause frame drops and other problems so the bottom end of the vertical capture area is clipped to end capture before Vsync starts.

PR #117, opened 2019-12-17
Support for new RGB/YUV board

IanSB:
Support for new RGB/YUV board:
Change YUV cpld to V3.1 to work with new pcb
Change RGB cpld to V7.1 to work with new pcb
Note v3.0 and v7.0 have been abandoned as they will not work properly with the new board so should not be included as cpld options in any release.

The RGB cpld has additional changes to support 4 level monochrome input:
Mux functionality has diverged and behaves differently from the standard BBC build:
When Mux is enabled, if Vsync is high, the two green bits are swapped.
This has the effect of making the vsync input an additional level sense on the green input:
Normally only 3 levels are detected, 0%, 50% and 100% as 00, 01 and 11
When mux is enabled, the Y/G sync should be set at a level between the lo and hi Y/G levels
This means that Y/G starts at 00 then 01 then when the Y/G sync level is exceeded the bits are swapped to 10 then finally 11 when Y/G hi is exceeded
Note this bit swapping will have no effect on any of the existing 3 level palettes (Amstrad/Spectrum) as 00 and 11 are still 00 and 11 when swapped and both 01 and 10 are set to produce the same colour.

To use the above you must connect the green and sync lines together and set the sync level and then the three Y/G levels and select the 4 level mono palette.
I have tested this with the Atom's monochrome output and it works perfectly with black, white and 2 grey levels being displayed. (See Atom mono profile in RGB Analog profiles)

There is also palette support for 6 level mono sources by connecting the green, red and sync signals together with termination disabled and using an external 75R terminating resistor (to avoid double termination)
The red input provides two more comparators for grey level sensing.  

Other changes:
Noise on the sync inputs caused crashes which was due to the nlines and vsync measurements producing garbage timings so they are now sanity checked.
Vsync detection in separate H/V mode has changed to work with both normal Vsync from CGA/VGA etc and to also extract vsync from composite sync when using 6847 type sources.


IanSB:
I removed that Atom/6847 logical palette generation from the YUV cpld so that it can work with other palettes. The 6847 Y/PA/PB signals are now translated to a palette in the Pi, same as for the normal cpld.

Note: for now I have left the signal conditioning that makes the RGB/YUV board look like the Atom board in the YUV cpld in case changing it messes up the noise reduction algorthm as the state of some signals would be inverted.

The signal conditioning is:
AL and BL are inverted and both LH and LL are changed to invert the video if using direct 6847 connection. Can you check to see if these are necessary for your noise reduction.
(They are done in the Pi on the RGB cpld)

I have also got the TI-994/a working but it's very experimental at the moment and not good quality:
capture4
The PAL version has a direct YUV output for an external PAL coder so it's simple to connect but it needs constant adjustment and varying some of the comparator levels by +-1 will get noise. The NTSC version has composite output so wouldn't work.

There seems to be some ripple noise like a 50Hz hum in the signal but the supply levels are clean and stable so not sure where it is coming from. I've tried with the back porch clamp and it seems to help the luminance but I think the R-Y and B-Y signals also need clamping.

Also there aren't enough comparators to determine all 16 YUV combinations and as a result I can't detect the difference between light grey and white so they both come out as white.
Not really sure where to go with this one, try to adapt the existing board for it or look at an alternative design. Supporting this would also add support for all MSX 1 computers which used the same video chip.

hoglet67:
>  Can you check to see if these are necessary for your noise reduction.

By noise reduction, you mean these expressions?
```
 AL <= (AL1 AND AL2) OR (AL1 AND AL3) OR (AL2 AND AL3)
```
All that's happening here is that three successive samples are being considered, and the output is effectively a majority vote of the three bits.
```
000 => 0
001 => 0
010 => 0
011 => 1
100 => 0
101 => 1
110 => 1
111 => 1
```
The should work fine if the inputs are inverted.

(Is that what you were asking?)

BTW, got my RS/Farnell orders today. Hopefully I'll get the board you sent tomorrow.

Dave

IanSB:
> Is that what you were asking?

Yes, I've updated the YUV cpld to remove the inversions and it now handles 6 bit data the same as the RGB cpld. It also handles all forms of sync like the rgb cpld (an earlier update)
One thing it doesn't have is programmable delay which might be required to get alignment for any proposed 6 bit in band palette control.
BTW Boards posted on Monday afternoon 1st class.

hoglet67:
Ian,

Just getting back to this after a diversion into the land of Javascript/Emscripten.

The additional Info on the Calibration Screen is great.

I've merged everything, plus a few fixups for errors I spotted.

Can you review these and make sure you are happy - most are obvious.

With the VHDL fixups:

1. the formatting has go a bit screwed up with tabs

2. VERSION_NUM_BBC was still 6.5, so I assume the intent was that when SupportsAnalog is false, the fusemap is identical. But that doesn't seem to be the case, for 2 reasons:
a) there were certain cases where the analog in/out was being driven when SupportAnalog=0
b) monochrome on vsync (added in 6.5) was dropped (so actually this is more like 6.4)

3. To address 2a, I've added an if/generate at the bottom, and to address 2b I've incremented the revision to 6.6 and generated a new version.

(The jedec fusemap from the 7.1 build is unchanged, so no need to refresh that)

This all took longer than I had planned, so I haven't had a change to build the new analog board yet. Hopefully tomorrow....

Dave

Issue #119, opened 2019-12-22
Future changes to RGB/YUV interface


IanSB:
There are some potential changes to the next issue of the RGB/YUV interface:
1. Move the termination control from the DAC output to the CLKEN line. The CLKEN GPIO signal is used during updates of the CPLD register chain and the DAC plus reprogramming the CPLD but at other times its state is irrelevant so it could be used to control termination on/off. This means termination will 'blip' during CPLD and DAC updates but that could be mitigated by feeding it through an R/C network to filter out the blip and only leave the long term logic level.
This would free up one DAC to allow a fourth level for BR / UV comparators which might help in the case of the TI-994/a and allow 4 levels of RGB as there are already 4 levels available on the green channel. The extra hardware support required for this would be 2 comparators and probably 1 TTL logic IC to do the 4 level to 2 bit  encoding which wouldn't fit on the current board without moving to SOP packages and probably a 4 layer PCB so would not be implemented on the next issue of the current board but any future board with this would be software compatible with the current design if this change was made now. One minor problem I noticed when testing this is that the CLKEN GPIO line is pulled high for several seconds during Pi bootup which is undesirable as it means unterminated inputs get a termination load during that time. That could be fixed by using normally closed analog switches and inverting the switching and clamping logic.

2. Add clamping to the BR / UV inputs. This could be done in two ways: Add another switch IC and associated passives or reconfigure the existing termination/clamping design to have 3 clamping switches and 1 termination switch instead of 1 clamping and 3 termination. 
The former would again not fit on the current board space without moving to SOP/4layer but the latter design would likely fit although it would mean a change in the termination functionality so that all 3 inputs would normally be terminated with a weak pulldown which was just enough to allow discrimination of the three RGB levels on the Amstrad and Spectrum and only the green/Y input could be terminated in 75R which would mean that could only be used in mono or maybe YUV mode.

IanSB:
#1 was addressed with the issue 3 analog board
#2 was not implemented

Issue #118, opened 2019-12-21
Seven bit mode


IanSB:
It should be possible to tweak this project:
https://github.com/c0pperdragon/A-VideoBoard
to output four or seven bit TTL for C64 and Atari respectively (C64 has 16 and Atari has 128 colour possible combinations).
This should be fine for the C64 but currently we are 1 bit short for Atari machines.

We should be able to get a 7 bit mode running without any hardware changes by using the VSYNC input as the seventh bit. (The 7 bits will fit into the 8 bit frame buffer without any problems as we only use 1 bit for the menu).

To do this without hardware changes would mean that only 1 seven bit pixel could be acquired per psync cycle as there are only 12 GPIO pins to transfer data which would halve the max pixel clock rate but as it currently maxes out at 28Mhz in six bit mode that would mean 14 Mhz in seven bit mode and the Atari has a ~7.16Mhz dot clock so there should be plenty of headroom.

There are a couple of registers left in the RGB CPLD so enough to add 1 bit to the capture mode setting but I don't know if the remaining logic could be squeezed into it.
In any case if it proves impossible, a new 7 bit CPLD could be designed.

hoglet67:
>  BTW do you think it would be possible to get a 7 bit mode implemented in the RGB cpld?
( see #118 )

I tried a simple experiment - just extend the scan chain by 1 bit (i.e. 1 extra register), and this failed to route (even with this exhaustive fit mode enabled).

The only way I could get it to route was ti unconstrain the pinout (by removing the .ucf file)

So I think it's unlikely this will ever fit unless the existing logic can be somehow simplified.

One thing I meant to ask you - one of the recent changes (swap_bits) has made the logic for the G functions a bit more complex. Can you explain what this is actually doing?

If you want to pursue this, I would suggest starting a new CPLD design that strips out the multiple sample offsets. That frees up a big chunk of space. As far as I'm aware, only Beeb Mode 7 has such an asymmetric pixel clock.

For future reference, here's the diff of what I tried:
```
diff --git a/vhdl/RGBtoHDMI.vhdl b/vhdl/RGBtoHDMI.vhdl
index 4c24da5..b45c3a6 100644
--- a/vhdl/RGBtoHDMI.vhdl
+++ b/vhdl/RGBtoHDMI.vhdl
@@ -68,7 +68,7 @@ architecture Behavorial of RGBtoHDMI is
[...]

PR #122, opened 2019-12-23
Mode 7 scaling bug fix, calibration summary split & hide pixel sampling 

IanSB:
A few minor but important changes:

Fix a long standing bug with mode 7 uneven scaling: It should have been scaling by 2.6666' : 2 (will be displayed as 2.67 : 2) but due to rounding errors it was scaling 2.63 : 2 which meant it was a bit more uneven than it needed to be. (I noticed this because of the new source summary info but had suspected there was a problem for some time due to the bad apple demo).

There will still be some unevenness in characters but separated graphics should now be fully even. The best way to see this is to view the bad apple demo before and after. The before version will show a few vertical black lines across the screen in the separated graphics that were wider than others. With the new version they are all even.

The calibration menu has been split into source summary and calibration summary.

The pixel sampling option is now hidden in rgb/yuv modes (could be extended to be hidden in non v1/v2 modes).

Also the Atom MKI palettes have been renamed to Dragon/CoCo/Atom

CLKEN now follows the termination setting to allow experimentation with freeing up one DAC.
(  see https://github.com/hoglet67/RGBtoHDMI/issues/119  )

BTW do you think it would be possible to get a 7 bit mode implemented in the RGB cpld?
(  see https://github.com/hoglet67/RGBtoHDMI/issues/118  )


IanSB:
In addition to the mode7 even/uneven scaling, I've now added a normal even/uneven scaling option which does the same for non-mode 7 modes.
This option is useful for the Oric and Camputers Lynx as they use 6Mhz dot clocks and any modes that use 6 or 12 Mhz dot clocks (like mode 7) don't scale nicely to 4:3.
The Oric can work around the problem by scaling like an NTSC source as it only has 216 lines vertically (similar to the current Atom profiles) without the PAL "letterboxing"
Mode 7 and the Lynx can't work around it as they have ~256 lines vertically.

PR #125, opened 2020-01-04
User definable palettes & v7.0 YUV cpld with adjustable clamp

IanSB:
User definable palettes have been implemented as follows:
On startup the software checks for a Palettes folder plus the default palette files and if they don't exist it creates them.

It then enumerates the Palettes folder and adds any files to the palettes option in the menu.
This allows extra palettes to be added and also existing palettes to be modified as they won't get overwritten. The palettes are currently binary and 1K in length (256 x 32 bit words)

The data for each 32 bit word is stored as follows:
mono << 24 | blue << 16 | green <<8 | red
Mono is the value to be displayed in mono mode. This is normally converted from the RGB values using an RGB to YUV algorithm but for the 6847 palettes it is the original 4 level mono irrespective of the RGB values.

Also YUV v7.0 with adjustable clamp:
The options are:
Sync tip
Back Porch short
Back Porch medium
Back Porch long

This uses a 2 bit counter triggered from the trailing edge of hsync.
I figured out what was wrong with the V6.0 cpld:
Under certain conditions the counter will wrap around at 6 bits:

> counter(5 downto 0) <= counter(5 downto 0) + 1;

but the clamp end value assumes a 10 bit count so under certain conditions it never gets to the end.

I worked around this with the variable back porch by limiting the reading of the counter to those 6 bits but it means that the "Back Porch Short" setting is very short at approx 600nS. Medium is ~2.8uS and long is 5.1uS.
Can you look at the counter to see if the 6 bit limitation can be changed.

I also added a confirmation to the CPLD update.


IanSB:
I've also added support for V1 and V2 profiles although this may need some further investigation as the timing for v2.0, through v2.4 differ so the division may be finer than just a V1/V2 split
I've setup the default timings for V1 and V2.4 but you really need to investigate what CPLD versions are still in the field as those are the ones that need to be supported.

IanSB:
After experimenting further, I've changed the way V1 and V2 profiles are handled:
Now there is only 1 folder called "OBBC" (for Old BBC cpld) and this contains all the profiles that are needed to support all the V1 and V2 sub-versions. 

They are:
BBC_Micro_v10-v20
BBC_Micro_v21-v23
BBC_Micro_v24

As you can see, the split is not across V1/V2 boundaries so you will need to select the appropriate profile after reprogramming the cpld.

I also included the V1 and V2 Atom profiles but I can't test them so can you check them and which versions they run under, renaming and adding any missing ones if required.

V30 cplds onwards seem to work OK with the standard profiles.

Note if testing the old CPLDs on the new 6 bit boards you must tie the Vsync input to +5v to replicate the way it was on the 3 bit board. If you don't do this, the sync will get inverted on some older versions and those cplds won't work at all.

IanSB:
I've just changed the following lines in the YUV CPLD v7.0:

Line 246
from:   counter(5 downto 0) <= counter(5 downto 0) + 1;
to:       counter(6 downto 0) <= counter(6 downto 0) + 1;
and Line 335
from:   if counter(6 downto 0) = "01" & offset then
to:       if counter(6 downto 0) = "11" & offset then

The line 335 change increases the length of the clamp pulses but without the line 246 change the clamp gets stuck on at max length.

The line 246 change appears to work and gives clamp lengths of about 1.7uS, 3.9uS and 6.2uS on the dragon and about 1.6uS, 3.8uS and 6.1uS on the Spectrum.
These values seem to be quite optimal as the middle one is suitable for the above ~7Mhz clock computers with the shorter one suitable for lower clock rates (as it will increase) and the longer one suitable for higher clock rates (as it will decrease) 

I haven't committed this yet but does changing the line 246 counter cause any potential problems as I'm not sure what you are doing there?

hoglet67:
> I haven't committed this yet but does changing the line 246 counter cause any potential problems as I'm not sure what you are doing there?

I can't see this causing any problems.

The counter serves a number of purposes:
- bits 3..0 control the sampling and loading the quad
- bit 4 controls the phase of psync
- the MSB (currently bit 9) controls whether PSync is active

Pixel sampling within a line starts as soon as the MSB of Psync becomes a '0'. It's important that the MSB doesn't become a '1' during the active part of the line. This is why the addition in line 246 is truncated.

It can safely be increased from (5 downto 0) to (6 downto 0), as this still would impact the MSB. It may, however, consume more product terms.

Dave

IanSB:
> It may, however, consume more product terms.

Definitely hitting a limit now. I just made another change which increased the clamping times a little further and they are now 2uS, 4.2uS and 6.4uS on the dragon.

I did this by changing the second line above to:
if counter(6 downto 0) = ("00" & offset) - 16 then
which tries to make more use of the full range of the 7 bit counter.
(Having to include the offset effectively subtracts those 5 bits from the full range and if the offset wasn't included then the clamp pulse would vary with the offset value)

Ideally to get the full range the above should be:
if counter(6 downto 0) = ("00" & offset) - 1 then
but that won't fit as I guess a 7 bit adder requires far too many product terms.

However I found the -16 variant fitted and that at least regained half of the wasted time as that only required a 3 bit adder.
Do you have any other ideas to improve this?

hoglet67:
> Do you have any other ideas to improve this?
```
counter(6 downto 0) = ("00" & offset) - 1
```
The -1 (or -16) adds a lot of complexity to this.

Wouldn't 
```
counter(6 downto 0) = ("00" & offset)
```
work just as well?

The difference to the clamp output is that it ends 20ns later.

Also, it might conceivably be possible to use counter(8 downto 7) as the clamp counter, saving two macrocells. Not sure if that would fit any better though, as both macrocells and product terms are scarce.

Dave

IanSB:
> Wouldn't counter(6 downto 0) = ("00" & offset) work just as well?

No, because with trailing edge sync the condition is immediately true and you get a zero length short blanking pulse with med and long reduced by an equivalent amount. This does work with leading edge sync because the leading and trailing edges aren't in phase.

What should work is delaying the hsync by another clock cycle to get HS4 and using HS3 and HS4 to reset the clamp counter while the main counter is loaded using HS2 and HS3 so it will be one count ahead.
However changing to ("00" & offset)  and adding HS4 as another delayed hsync doesn't fit either. 
 I'm surprised at that because it didn't seem to be adding that much complexity. Perhaps you can re-arrange things to get that to fit.

Note as a general rule, using leading edge sync will result in clamp pulses that are shorter than trailing edge sync due to the out of phase triggers for the counters. In a worst case that could be a significant amount (nearly the entire length of the short blanking pulse) but for the spectrum which is the only leading edge use at the moment it is ony 0.1uS or so.

hoglet67:
> However changing to ("00" & offset) and adding HS4 as another delayed hsync doesn't fit either.
> I'm surprised at that because it didn't seem to be adding that much complexity. Perhaps you can re-arrange things to get that to fit.

The problem is once you get beyond ~90% product term usage, it gets very hard to route a design with a constrained pinout.

I've pipelined a couple of signals, and that's brought the product term count down by about 10% from
```
             64/72      157/216     325/360     29/34
```
to
```
             66/72      140/216     288/360     29/34
```
With this, it's possible to add HS4, which I've just done and pushed.

There may just about be space to implement a 3-bit clamping counter, which would give a bit more precision on the clamp duration.

Is that worth trying, or or you happy with were we currently are.

Dave

hoglet67:
One thing I noticed in your last batch of commits.... The code to remove inapplicable options from earlier versions of the YUV_CPLD got removed. I think that was because you re-ordered the options.

It's probably worth us doing this properly, so that options in the middle of a menu can be hidden if required. There were a couple of other cases where this would have been useful. It then removes any artificaly constraints on the order we choose to put menu items in.

hoglet67:
I've just pushed a v75 that is slightly more efficient than v74, by using a down counter for the clamp counter. This saves one macrocell.

I tried several ways to increase the clamp counter to 3-bit, but none of these fitted.

So I think I'm done for now.

IanSB:
>  I think that was because you re-ordered the options.

Yes, but it was also because it affects save and load of profiles as any profile saved/loaded with the truncated menu would also be truncated. The YUV one isn't really a problem as the versions so far have only been for testing and not been released so no one in the field is going to be using those versions.

> It's probably worth us doing this properly, so that options in the middle of a menu can be hidden

That would have to be implemented in a way that didn't affect the loading/saving of profiles otherwise you would end up with incompatible profile formats.

> There may just about be space to implement a 3-bit clamping counter

With 0.74 the clamp sizes are 2.23, 4.46 and 6.70uS with a 7.159090Mhz pixel clock (16, 32 & 48 clocks). With 3 bits that would be 8, 16, 24, 32, 40, 48 & 56 pixel clocks assuming a reduction of 1 bit in the sample counter (6 bits) or 16, 32, 48, 64, 80, 96 & 112 pixel clocks with no reduction (7 bits).
A couple of pixel clock speeds that I can think of that could be used with this are:

Apple II(c) is 14.31818Mhz which gives:
.558, 1.11, 1.68, 2.23, 2.79, 3.35, 3.91 with 6 bit sample counter
1.11, 2.24, 3.35, 4.46, 5.58, 6.70, 7.82 with 7 bit sample counter (as now)

Superboard II which is 4Mhz
2, 4, 6, 8, 10, 12, 14 with 6 bit sample counter
4, 8, 12, 16, 20, 24, 28 with 7 bit sample counter

Neither of these options are ideal and perhaps the best option would be a 6 and a half bit counter (i.e. 96 clocks instead of 64 or 128). Another option might be to split the clamp settings between the 6 and 7 bit counters but those options will probably intruduce too much complexity so I think the most useful of the above two would be a 3 bit clamp counter with the existing 7 bit sample counter although the highest couple of settings are unlikely to ever be useful.

That would give 4.46uS clamp pulses with  7.159090Mhz and 14.31818Mhz and 4uS with 4Mhz and 8Mhz. I could probably add an Auto option to the clamp setting that picked the value that produces the closest pulse to 4uS.

hoglet67:
The CPLD is pretty much full now, I think you will struggle to add anything else.

Re: the clamp duration, I would expect that a shorter clamp pulse would still work, it might just take a bit longer to settle. I wonder if there is any way we could test this?

IanSB:
> The CPLD is pretty much full now

Even with just the 3 settings you get usable results in the ~ 3uS - 5uS range for all the pixel clocks I'm aware of so far: (You can't use much above 5uS as it will start to clamp any border colour in active video)

16000000 = 1.00, 2.00, 3.00
14318180 = 1.12, 2.23, 3.35
10738635 = 1.49, 2.98, 4.47
 8000000 = 2.00, 4.00, 6.00
 7159090 = 2.23, 4.47, 6.70
 7119090 = 2.25, 4.49, 6.74
 7000000 = 2.29, 4.57, 6.86
 6500000 = 2.46, 4.92, 7.38 
 5369317 = 2.98, 5.96, 8.94 
 4000000 = 4.00, 8.00, 12.00
 3579545 = 4.47, 8.94, 13.41

> I would expect that a shorter clamp pulse would still work,

I have tried this and the short pulse (~2uS) works with the Spectrum but I have a problem with the Dragon as there seems to be mains hum coming from somewhere and that can be seen when using the short clamp as a noise bar ripples through the screen. I recapped the psu but it didn't help.
I'm going to change the way I pick up the mono video. At the moment it comes straight from the 6847 which is low level but I'm going to try the video output instead but I'll have to disconnect the colour subcarrier from it as well.

BTW do you have a 50Hz colour board for your Atom? If so can you tweak the Atom 50Hz TTL profiles for V1/V2 and standard BBC/RGB profiles as I managed to get them setup approximately but they need a real source to get them finalised.
I think you also have a superboard II so maybe you can create a mono profile for that as well.

hoglet67:
> I'm going to change the way I pick up the mono video. At the moment it comes straight from the 6847 which is low level but I'm going to try the video output instead but I'll have to disconnect the colour subcarrier from it as well.

I have a Dragon motherboard that I used for testing the ICE-6809E. Unfornately, I don't have a keyboard. But I can use the ICE to write to screen memory. Is it worth me setting this up to see if I also have an issue with mains hum?

> BTW do you have a 50Hz colour board for your Atom? If so can you tweak the Atom 50Hz TTL profiles for V1/V2 and standard BBC/RGB profiles as I managed to get them setup approximately but they need a real source to get them finalised.

I do have a 50Hz Acorn colour board, but I've never used it, and it's not currently installed. The Atom profiles that previously existed were used with one of Phill's Colour boards, which outpus TTL sRGB at 60Hz.

> I think you also have a superboard II so maybe you can create a mono profile for that as well.

Yes, I've just added a profile for that.

It's quite tricky to get the geometry right, because it can render characters in the overscan area. The screen memory is 32x32 chars. Nominally 24x24 is visible, but this would depend on your TV. In fact, there is no guaranteed back porch, so I ended up using the sync tip clamp.

IanSB:
>  Is it worth me setting this up to see if I also have an issue with mains hum?

Do you have the standard dragon PSU & regulator?

> I do have a 50Hz Acorn colour board, but I've never used it, and it's not currently installed. The Atom profiles that previously existed were used with one of Phill's Colour boards, which outputs TTL sRGB at 60Hz.

That explains why the previous TTL profiles looked like 60Hz

I think the only difference between 50 and 60Hz profiles will be the vertical offset and the total number of lines settings.

> It's quite tricky to get the geometry right, because it can render characters in the overscan area.

Yes it's the same for the UK101. The discrete video generation circuit uses the simplest design that doesnt account for blanking and relies on those character positions to be cleared to spaces. (There are even characters occupying the sync period). I set the minimum area to be the area used by BASIC to print text using the default terminal width and then set the maximum area to be the largest sensible size possible.

> so I ended up using the sync tip clamp.

Probably a good idea as games used to move characters into the blanking areas.

Is your superboard 50 or 60Hz? (I think some were modded to 50Hz) and how many lines does it generate (I noticed you put 256 in the geometry instead of the typical 262 or 312

BTW what is the superboard x2 version for?

There is one problem with the UK101 and it probably affects the superboard as well:
The H and V sync pulse positions and durations are set by RC monostables and as a result their relative timings vary with temperature/time and it sometimes requires the vsync setting to be switched to Electron type to get a judder free image. As this would require continual adjustment, I'm currently looking at trying to auto detect the sync state to avoid the need for the standard/electron vsync switch. 

Do you have any ideas about that?

hoglet67:
> Do you have the standard dragon PSU & regulator?

No, so I would have to run it off my bench power supply.

Maybe there's not much point in this, as the power supply is the most likely source of 50Hz.

> Is your superboard 50 or 60Hz? (I think some were modded to 50Hz) and how many lines does it generate (I noticed you put 256 in the geometry instead of the typical 262 or 312

It's 60Hz.

The crystal is 3.932160 MHz (60*256*256).

So there are just 256 lines, with each line being 256 clocks (65.104us)

> BTW what is the superboard x2 version for?

Each pixel is on for 1/2 a clock cycle, and off for 1/2 a clock cycle.
capture0

This is visible if I connect the composite video directly to my TV/Monitor.

The x2 profile makes this visible on HDMI by sampling each pixel twice (using a 2x pixel clock)

Also, using a 2x pixel clock reduces the offset between the trailing edge of sync and the start of PSYNC. In the YUV CPLD this is currently 256/8 pixel clocks.  With the 1x pixel clock, this is 4 characters, which means you can't get much overscan. With the 2x pixel clock, this reduces to 2 characters, so you can get a 28 character wide display if you want.

I wonder if we should get rid of this delay in the CPLD, like we did with the RGB CPLD?

> The H and V sync pulse positions and durations are set by RC monostables and as a result their relative timings vary with temperature/time and it sometimes requires the vsync setting to be switched to Electron type to get a judder free image. As this would require continual adjustment, I'm currently looking at trying to auto detect the sync state to avoid the need for the standard/electron vsync switch.
> 
> Do you have any ideas about that?

I haven't noticed any instability.

Would using the leading edge of VSYNC as the reference help in this case? (I assume this sync state delection would be happening in software).

IanSB:
> Each pixel is on for 1/2 a clock cycle, and off for 1/2 a clock cycle.

Interesting, I hadn't realised the superboard did that (the UK101 doesn't)

> I wonder if we should get rid of this delay in the CPLD, like we did with the RGB CPLD?

Probably a good idea, I hadn't realised it was still in there.

> I haven't noticed any instability.

You may just be lucky with the range you have, I put some pots on my UK101 monostables a long time ago so I could adjust the picture position and adjusting those can make the problem come and go but I also found that adjusting them to a 'good' position when cold and leaving it to warm up resulted in it drifting into 'bad' mode.

hoglet67:
> Probably a good idea, I hadn't realised it was still in there.

It comes from these two constant values:
```
     -- Default offset to start sampling at when using the leading edge of sync
    constant leading_offset   : unsigned(9 downto 0) := to_unsigned(1024 - 512, 10);
    -- Default offset to start sampling at when using the trailing edge of sync
    constant trailing_offset   : unsigned(9 downto 0) := to_unsigned(1024 - 256, 10);
```
If it wasn't for the need to measure the sync pulse width (to reset the phase of the PAL inversion), the counter could be reduced in width.

Regarding the PAL inversion, do you see this being used for anything other than the Spectrum?  It's only going to work for pixel clocks upto about 12MHz.

Also, on the Spectrum when you change from leading to training edge there is very little shift in offset. That will only work like this if the pixel clock is ~7MHz.

Dave

IanSB:
> Regarding the PAL inversion, do you see this being used for anything other than the Spectrum? It's only going to work for pixel clocks upto about 12MHz. Also, on the Spectrum when you change from leading to training edge there is very little shift in offset. That will only work like this if the pixel clock is ~7MHz.

So far the only machine that requires PAL inversion and leading edge is the Spectrum and I don't know of anything else that would require it.

I think all the other computers I've looked at are either mono which should work with the interface or composite on chip which won't.

hoglet67:
> > It's probably worth us doing this properly, so that options in the middle of a menu can be hidden
> 
> That would have to be implemented in a way that didn't affect the loading/saving of profiles otherwise you would end up with incompatible profile formats.

I've just implemented a hidden attribute on properties that causes the property to be skipped when constructing the menus. This shouldn't impact the profile reading/writing.

4d76fce8

PR #128, opened 2020-01-09
Move sync type to geometry and add auto and non-interlaced options

IanSB:
VSync type has been moved to geometry as it really belongs with the other sync and line count settings and the existing Standard and Electron sync settings have been renamed as follows:

"Standard" to "Interlaced"
"Electron" to "Interlaced 160uS Vsync"

They should behave exactly as before but I've added two new options:

"Auto": which switches between the above two automatically.
"Non interlaced": which always forces detection of the same field and doesn't skip any lines.

The Electron actually has 160uS Vsync, instead of 128uS for 6845 type video (my issues post was incorrect about the Electron sync) and I used that to implement Auto switching by measuring the vsync period.

Most profiles now default to Auto but the UK101 and superboard default to non-interlaced.

The Spectrum palette regression should be fixed - can you confirm

I've also simplified the screencap generation code and tweaked it so that non-full-screen screencaps always have the correct aspect ratio just like full screen ones.
 
Normally standard screencaps are half scale (so 800x600 for a 1600x1200 monitor) but some have to be full screen to get the correct aspect ratio such as the superboard profiles so their screen caps will always be large.
Also mode 7 has been tweaked to produce a 3x2 aspect ratio rather than 1x1 so it is now too wide instead of too narrow and the default screencap is 1800x1200 instead of 600x600.

Not sure if these larger screencaps are a good idea but this can easliy be reverted - let me know which you prefer.

The non-interlaced setting still suffers from problems when the end of vsync merges with the first hsync so further work needs to be done there but I think I'm going to call a halt at this point, to work on the documentation and get a release ready so any further changes for now will just be bug fixes.

Work on non-interlaced and the in band palette code can be done after a beta has been released.


IanSB:
I tweaked mode7 screencap a little more as it was producing default caps that were larger than the screen (1800x1200 on a 1600x1200 monitor):

For non-scaled screencaps on a 1600x1200 monitor: 
Uneven scaling produces 600x600 (1:1)
Even scaling produces 1599x1200 (3:2)

For scaled screencaps on a 1600x1200 monitor
Uneven scaling produces 1200x1200 (1:1)
Even scaling produces 1599x1200 (3:2)

So that covers all practical options.

PR #135, opened 2020-02-02
Multi Pi support, video type option, hide offsets & system summary menu

IanSB:
These commits add multiple Pi version support at any frequency although it defaults to standard speeds for each model. Edit the default.config file + config file to change frequencies.
It seems to work reasonably well with the Pi1 and Pi3 after I made some changes to the capture code including adding a second core to do the capture on the Pi3 for advanced deinterlace and that helped a lot but it still has the occasional glitch on that model. A reboot option has been added to the Info menu so models without the reset header can be restarted.
It's untested on the Pi2 as I don't have one but I suspect it will still be slightly glitchy in mode 7 without overclocking.

Also a new video type option has been added in the Geometry menu which has options of "Progressive" and "Interlaced Teletext". This change allows teletext modes to be created for other than mode 7. This is primarily to support the Acorn system 40 column teletext card (I just received a blank PCB for one so will be building it up sometime).

Previously the mode7 setting was used to switch between mode7 and normal configs and also to configure the capture code for teletext. These two functions have now been split with mode7 just controlling the switch between mode7 and normal configs with the new video type option controlling the switch between teletext and non-teletext capture code. Some options in the preferences menu have been renamed to reflect this.

This means you can now have teletext in the normal profile as well as the mode7 profile and I have created a "Teletext" profile to test this.
Note if you select the teletext video type when not using 3 bit capture with a 4 bit buffer you will get a mess on screen as there are no capture loops for 6 bit capture  / 8bit buffer. 

The individual offsets (A, B, C etc) in the sampling menu are now hidden in all modes other than mode7 to reduce the size of the sampling menu most of the time.
 
The firmware version menu has been renamed to System summary and displays clock frequencies in addition to firmware version and temps/voltages.

I think this is a potential release candidate for for the next version as we really should get a release out there in the master branch as it's confusing people.

Please test this with as many combinations of profiles and CPLD versions as you can and I will do likewise.


hoglet67:
> Please test this with as many combinations of profiles and CPLD versions as you can and I will do likewise.

I've been holding back on testing and making any documentation changes as things have been changing very rapidly.

I'm happy to do some testing over the next week.

Are we agreed then that we'll put a freeze on adding new features until after we have merged back to master?

Dave

hoglet67:
Hi Ian,

I've just tried my Pi 3A+ and hit a couple of issues.

1. The firmware blob we have (2019/04/19) doesn't support this Pi. So I've just dropped a later version onto the card (2020/01/16) and this boots fine now.

2. Mode 7 (12MHz) works perfectly, but Mode 0-6 (16MHz) is suffering from massive jitter. This is running with the CPU Clock of 1400MHz and a CORE clock of 400MHz.

capture4

capture2

Is this consistent with what you are seeing, or am I doing something wrong?

(This is with the original 3-bit board, with the latest (v6.6) CPLD)

I've also tried my Pi 2 and hit some different issues:

1. Mode 7 detection goes wrong, and it spends all it's time switching modes

2. Mode 0-6 gives a stable display (with none of the jitter issues I'm seeing with the Pi 3A+), but the menu doesn't render at all. The screen dims and that's it.

hoglet67:
Looks like the Pi 2 issue was introduced in 9f543310

IanSB:
> It's a shame there are several independant things in this one commit. It makes it so much harder to narrow down on what's causing a regression.

Sorry about that, I never had a point where they were separately committable in a fully working form.

I haven't actually done any specific work on modes0-6 capture, all the changes were to mode 7 buy I saw the problem with massive jitter in modes0-6 on my Pi 3 when using 1600x1200 res but not when using 1080p res. I eventually found that reducing the max horizontal capture width from 768 to 720 got rid of it so I added silent clipping to 704 in geometry.c:

`#if defined(RPI2) || defined(RPI3) //clip capture size to avoid glitches on RPI2 & RPI3 (not needed on RPI4)`

(This also reduces the capture height to keep a roughly 4:3 capture area)

I think when we last tested PI2 & PI3 we were only capturing 672x270 and now the max is 768x288 which is one reason that might have not been an issue before. Mode 7 has a similar increase in capture width but that is not clipped.

Try reducing the clock speed of the Pi3A+ to 1200 Mhz in config.txt to see if that makes any difference as that should be equivalent to the speed I'm using on my Pi3. Also try reducing the max horizontal capture to less than 704

Other changes were switching from PLLC to PLLA and adding the core in mode7 advanced. 
You may have noticed I added a memory benchmark and I saw in early testing that running a second core resulted in a significant increase in uncached memory access times. This currently doesn't show up in the benchmark as the core is sleeping during the benchmark and only woken up for part of each line during mode 7 advanced line capture but if you remove the wfe instruction in the run_core code you should notice the difference. BTW Uncached memory is slower on the Pi3 than the Pi0.

IanSB:
> Are we agreed then that we'll put a freeze on adding new features until after we have merged back to master?

Yes, the only reason I added the teletext change was I could see it would be needed and I wanted to update the profile format to support it before going for a release.

BTW there is one issue I'm aware of with hiding the Offsets when not in mode 7: If the menu is on screen and the BBC mode is changed, the hiding/unhiding isn't updated to reflect the new mode. Perhaps you could look at that.

I've also updated the wiki with new screen grabs and text for the recent menu changes

hoglet67:
> Try reducing the clock speed of the Pi3A+ to 1200 Mhz in config.txt to see if that makes any difference as that should be equivalent to the speed I'm using on my Pi3.

OK, that alone fixed the issue I was seeing.

Which I guess is expected - as the only significant difference between the 3 and the 3A+ is the clock speed. I can see how having the ARM clock be an integer multiple of the Core clock might be advantageous in terms of memory latency.

> Also try reducing the max horizontal capture to less than 704

If I now change the capture width, I get some strange results.

What I'm seeing is bands of stability followed by bands of instability.  Specifically, the following capture widths exhibit jitter:
- 248-288
- 488-544
- 632
- 648-672
- 760-800

(I commented out the slient clipping for this test)

There doesn't seem to be any obvious pattern here.

I'm not convinced that silent clipping at 704 is helpful. If I set overscan to half way, I get a capture width of (640+704)/2 = 672 which lands in one of the dodgy regions. Without silent clipping, you get (640+768)/2 = 704 which is fine. So while there may be some cases where it helps, there are other cases where it doesn't.

I would definitely suggest getting rid of this for now, until we understand this effect better.

Dave

hoglet67:
Another data point regarding the MODE 0-6 jitter on the Pi 3 family.

If I set the FB size to Normal (so line doubling is disabled) then the jitter goes away at all capture widths.

This is consistent with what was seen when I first tried to get this to work on the Pi 3 - see commit 
34a9cafe

At the time I tried quite hard to fix this, re-writing the capture code in lots of different ways, but to no avail. At least now we have the option to run in FB Size=Normal and avoid the need for line doubling.

IanSB:
> some strange results

I think that just about sums up the Pi 2/3 memory system

>see commit 34a9caf

I guess that got lost when I added the menu driven double height option.

I'll remove the silent clipping and add double height forced to off on Pi2 and Pi3. The only consequences of that are scanlines won't work and vertical interpolation will be a bit softer.

hoglet67:
> I'll remove the silent clipping and add double height forced to off on Pi2 and Pi3. The only consequences of that are scanlines won't work and vertical interpolation will be a bit softer.

For some reason, I don't think the line doubling affects the Pi 2 in the same way. But I can double-check that tomorrow.

PR #136, opened 2020-02-04
Bug fixes

IanSB:
Some bug fixes following recent testing.
Hold off merging for a few days as there will probably be some more and we can use this thread to discuss and collate them.

Could you test the Pi2 & 3 changes.

I've included the firmware I've been using which I downloaded on 23rd Jan. This includes the Pi4 firmware which is separate.

BTW I noticed you added Communicator profiles but that included a separate teletext profile. Does that mean auto switching doesn't work? Also does that teletext profile need to use the new teletext source type?


hoglet67:
> Could you test the Pi2 & 3 changes.

I'll do that today.

> BTW I noticed you added Communicator profiles but that included a separate teletext profile. Does that mean auto switching doesn't work?

That's correct, it didn't work. More investigation is required, we didn't have access to a scope at the time.

> Also does that teletext profile need to use the new teletext source type?

The communicator uses a different teletext chip (SAA5240A) that doesn't use character rounding. So the font is very different, and the data in the odd and even fields is the same. So I don't think in needs to be treated as a teletext source, as no de-interlacing is required.

The datasheet is here:
https://www.datasheetarchive.com/SAA5240-datasheet.html

Dave

IanSB:
I noticed that the GPIOs can have software controlled internal pullups which means the pullup resistors on the switches aren't really needed, any reason why you didn't use them?

hoglet67:
> I noticed that the GPIOs can have software controlled internal pullups which means the pullup resistors on the switches aren't really needed, any reason why you didn't use them?

No good reason now.

In the first version (on a breadboard) the switches were routed via the CPLD, and that needed pullups. When the CPLD was bypassed, I never got rid of the pullups,  because I wasn't particularly pushed for space.

I think the Pi pullups are fairly weak - about 50K - but this should be fine for a switch.

hoglet67:
From the earlier thread...

> BTW there is one issue I'm aware of with hiding the Offsets when not in mode 7: If the menu is on screen and the BBC mode is changed, the hiding/unhiding isn't updated to reflect the new mode. Perhaps you could look at that.

I've just fixed that - the only complex part was keeping the cursor in the same logical place, when other params further up are added/removed.

> Could you test the Pi2 & 3 changes.

No issues so far on the Pi 2 or Pi 3A.

A couple of thoughts....

1. Should we update the config.txt and default_config.txt for the Pi 3+ so that it runs at 1200MHz?

2. On the Pi 2/3, it's still possible to select Double Height in the UI, it's just silently ignored, which has the potential for user confusion. I'm wondering whether we're doing the right thing here by supressing this option. I'm guessing the jitter is only a problem if the pixel clock is 16MHz. For machines with a much slower pixel clock, we are probably needlessly suppressing this option.

I guess this comes down to whether we plan to publically support the Pi 2/3 models. It's already consuming quite alot of our bandwidth, and any time spent testing on the Pi 2/3 is time not spent testing on the Pi Zero.

To be honest, I'd rather be focussing my testing time on the Pi Zero.

Dave

IanSB:
> Should we update the config.txt and default_config.txt for the Pi 3+ so that it runs at 1200MHz?

Yes I was thinking that too to keep temps down. If you edit the config.txt & defaultconfig.txt I'll merge that back with your other changes

>On the Pi 2/3, it's still possible to select Double Height in the UI, it's just silently ignored, which has the potential for user confusion

The menu can always be tweaked to reflect the reality at a later date.
 
>I'm guessing the jitter is only a problem if the pixel clock is 16MHz. For machines with a much slower pixel clock, we are probably needlessly suppressing this option.

I tried that and the jitter is still present if you select double width as well which would usually be the case with the lower clock frequencies so you can either have double width or double height but not both under any clock frequency so I left it with double height always suppressed for now.

> To be honest, I'd rather be focussing my testing time on the Pi Zero.

Yes leave the Pi 2 & 3 for now, it seems to be usable if anyone tries it. The zero is always going to be the best option. 
I'll probably revisit the Pi4 later to see if it is usable at higher psync rates than the zero which has the potential to be useful.
 
Can you test on a standard beeb, mine is currently dead. I've tested on an M128, Electron, Amstrad, Oric, Lynx, and ZX80 with others to follow.

hoglet67:
> If you edit the config.txt & defaultconfig.txt I'll merge that back with your other changes

Done.

> Can you test on a standard beeb, mine is currently dead.

I'll do some more testing on the Model B tomorrow.

PR #137, opened 2020-02-05
Bug fix #2

### LINKED ISSUE #2: Periodically re-calibrate
e.g. Once an hour

IanSB:
Some more fixes and tweaks to profiles:

I previously renamed some palettes to fit on screen but forgot to change the names in the relevant profiles

I've removed the TI99/4a support for now as it doesn't really work yet.

The Spectrum +2 and +3 clocks were wrong, should be 7.093800 Mhz, not 7 Mhz (Uses PAL colour frequency * 8 as the master clock and that is the same freq / 5)

The 2 level palette is now only controlled by Y hi and the Y lo setting is ignored, previously both had an effect which was confusing.

You may need to revisit the superboard profiles as they didn't have termination on. I've enabled it but that might have an effect on the sync and Yhi levels. Also they were setup for Sync tip clamping. Would it be better to use back porch? (If so the clamp level will have to be increased from 0)

I've tested Atom dedicated, Atom RGB/YUV, Spectrum +3, Tandy CoCo, Dragon, Spectrum 48K, UK101 composite and ZX81 ULA output today.

Should the Input Mux be renamed to something like 4 Level G & 4 level Y in the RGB and YUV cplds respectively? That's what it does with those cplds but it still works as  Input Mux in the BBC cpld.


hoglet67:
> You may need to revisit the superboard profiles as they didn't have termination on. I've enabled it but that might have an effect on the sync and Yhi levels. Also they were setup for Sync tip clamping. Would it be better to use back porch? (If so the clamp level will have to be increased from 0)

In the manual, termination is not recommended, see page B5 where is says connect to the Hi Z input of the video monitor. This is because relatively high value resistors are used in the output circuitry (including a 5K variable resistor).

Here's a link to the manual:
http://www.osiweb.org/manuals/600revB.pdf

(The video schematic is on PDF page 53)

I also deliberately chose sync tip clamping, because there is no guarantee there will be a back porch (i.e it's possible to have character data displayed during the back porch).

I would suggest reverting back to the original version.

Dave

IanSB:
> I would suggest reverting back to the original version.

Ok I've done that and also tweaked the UK101 profile to match.

 I looked at the UK101's video out and it was 2v p-p unterminated and 700mv p-p when terminated, neither of which are correct but the sync to video ratio was something like 65 : 35 (the wrong way i.e huge sync pulses) so I've used the unterminated option which gives a better threshold range on the luminance (700mv p-p unterminated vs 250mv p-p terminated for the video part)

hoglet67:
Ian,

I've been testing on the Model B, with older CPLD. Could I have you thoughts on this issue?
https://github.com/hoglet67/RGBtoHDMI/issues/138

Dave

PR #141, opened 2020-02-06
Fix separate H & V sync issue and old cplds displayed when debug enabled 

IanSB:
There was a problem with switching between separate and composite sync. Previously in PC mode it used the detected sync type which caused it to get stuck in a no sync state if the detection was wrong due to syncs disappearing for a while during a PC mode change. Now it uses the expected sync type instead.

I also added some deglitching to the Vsync input (It reads the Vsync twice like the old firmware Csync).

Old cpld versions are now stored in "old" sub-folders and are only displayed when debug mode is on in the settings menu.


IanSB:
While doing some tests I accidentally programmed the 6 bit RGB cpld into a 3 bit BBC board and it destroyed the CPLD.  It got extremely hot and doesn't work anymore. (I have had a couple fail like that over the past year for other reasons and they seem to be quite sensitive to shorted outputs).

On the 3 bit BBC board GPIO22 and P19 are shorted together but on the new 6 bit board they are both used as outputs.

As this is a quite serious hazard I've made some changes to try and minimise the chances of this occurring:
The cpld type (ATOM / 3BIT / 6 BIT) is written to the volume label of the SD card during a successful bootup. If some sort of cpld programming failure occurs and it results in a blank cpld the volume name is used to determine which set of cplds to offer for programming (previously it always offered the 6 bit cplds)

This on its own still has issues because if you take an SD card from a 6 bit board and put it in a blank 3 bit board you will be offered the wrong cplds (and vice-versa).  You can swap SD cards between already programmed boards as the volume label will be updated on the first boot each time.

I then figured out a way to detect if the RGB cpld was programmed into the BBC board and it now fails safely without causing any damage (I hope!) This trick doesn't work with the YUV cpld although it might be possibe to implement something similar with a YUV cpld update.

This works as follows:
On the RGB cpld, P19 is used both as an input and an output. When the MUX and VERSION pins are set to 0 it is an input used to switch the cpld version ID between the analog and TTL versions. When MUX or VERSION are set to 1 it is an output with the clamp signal.

The STROBE_PIN (GPIO22) is set as an input on bootup (in case P19 is an output) then MUX and VERSION are set to 0 which will make P19 an input on the RGB cpld. Then STROBE_PIN is set as an output and set to 0 and the version ID is read. STROBE_PIN is then set to 1 and the version ID is read again. If the ID changes it means that the STROBE_PIN and P19 are connected together so the software fails with a fake cpld unprogrammed error and the BBC version is set as the volume label so the correct cplds are offered for reprogramming.

You can test this on a 6 bit board by connecting STB and P19 on the 4 way connector with a low value resistor (<100R) during bootup to simulate the 3 bit board. This will result in a cpld unprogrammed error and the BBC cplds will be listed for programming.

I've included a BLANK cpld file in the "old" folders and this can be used to erase the cpld to test bringing up a new board. (If the volume label is not ATOM/3BIT/6BIT then it defaults to 6 bit)

hoglet67:
Gosh, this is a tricky one!

Sorry for the long response...

> While doing some tests I accidentally programmed the 6 bit RGB cpld into a 3 bit BBC board and it destroyed the CPLD.

Am I right in thinking there are currently two ways to do this:
1. Start with a genuinely unprogrammed CPLD
2. Press all three buttons during boot

And in both cases the system defaults to the 6 Bit firmware?

Some observations....

1. I'm very surprised a short on a single output pin could destroy the CPLD. The CPLD output drive current is limited to ~32mA (output driving low, short to 3V3), or ~65mA (output driving high, short to 0V). Even in the second case, the extra power being dissipated is ~200mW.

A more common cause for CPLD destruction is something called "latch-up", which can cause the device to heat up very rapidly. The result is a CPLD where these is a virtual short between 3V3 and GND. If left for long-enough, the device is damaged, and this short seems to persist.

Could you measure the resistance between 3V3 and GND on your damaged device?

I'm not 100% clear on the causes of latchup in the XC9500XL family, and haven't experienced it myself. But I know Phill Harvey Smith (prime) destroyed many CPLDs while developing his Atom Colour Board. I believe the cause is exceeding the I/O voltage limits, which are:

> Input voltage relative to GND: 0.5V to 5.5V

But also:

> XC9500XL: You can drive into an unpowered I/O, provided Vin does not exceed VCCINT by more than 4.0V".

It's this latter case that we might be hitting.

During yesterday's event, was the board connected to / powered from a Beeb? And was that Beeb power cycled just prior to the CPLD blowing up?

2. We have some control over the Pi drive strength, so we might want to consider reducing the drive strength on this output, see:
https://www.scribd.com/doc/101830961/GPIO-Pads-Control2

(The document is clear that is not designed as a current limit, but it will have that effect)
 
3. I generally like your solution of trying to somehow test whether GPIO22 is directly connected to P19. I've looked over the schematics, and this is the only testable difference between the two boards.

With your current implementation, it seems that *any* difference in cpld version number between the first and second read will throw you "CPLD is unprogrammed" mode. I would code it to be much more specific. i.e. look for a specific change from  DESIGN_RGB_TTL to  DESIGN_RGB_ANALOG, and only in this case take evasive action.

The reason for suggesting this is that very occasionally I have seen unreliable CPLD version reads (resulting in an Unknown CPLD).

4. Making the CPLD Update more robust

Firstly, there is nothing on the Update CPLD Menu which states the currently inferred board type. We really should add this.

Why does the 3-button reset exist, and why, if the CPLD version was valid on a 3-button reset, is it ignored?

I can see there are development use cases for this function.

Do you see this as a last resort, if an end user has somehow managed to program the wrong type of CPLD,  there is some way to recover?

5. Writing the CPLD Type to the SD Card Volume ID

This actually makes me very nervous, because I move cards around all the time.

I think there is a better way to detect whether GPIO22 is connected to P19 or not.

JTAG supports commands to allow the external pins to be controlled, independantly of whether the CPLD is programmed or not.  The commands are called EXTEST PRELOAD and SAMPLE.

They allow the current state of the I/O pins to be read (or written).

This could be used to descriminate between the different boards we currently have.

The only difficulty is writing the C code exercise these commands, which would take some effort. So I'm not suggesting we necessarly do this now.

Dave

IanSB:
> A more common cause for CPLD destruction is something called "latch-up",

I think it was latchup, the cpld got very hot but it looks like reprogramming with the wrong cpld might have triggered it. Not absolutely sure but it was the one thing I did before I noticed it. I was using it with a bbc at the time and there was no power cycling until after I noticed the problem.

The previous one I had a problem with had one of the switch inputs (which are not used) shorted to ground internally and that got quite hot as well and eventually failed completely in a similar way.

BTW those unused three inputs could be used as another 3 bits for an 8 or 9 bit board albeit with lower max clock rate :)

> Could you measure the resistance between 3V3 and GND on your damaged device?

2.2 Ohms

> Am I right in thinking there are currently two ways to do this:
    Start with a genuinely unprogrammed CPLD
    Press all three buttons during boot

Or have a programming failure or program in the BLANK test file

> And in both cases the system defaults to the 6 Bit firmware? (before these changes)

Yes in all four cases

>With your current implementation, it seems that any difference in cpld version number between the first and second read will throw you "CPLD is unprogrammed" mode. I would code it to be much more specific. i.e. look for a specific change from DESIGN_RGB_TTL to DESIGN_RGB_ANALOG

OK I can change that. I left it like that as it was potentially possible to update the YUV cpld to behave in the same way and thus gain the same protection so I would have to test for all types.

> Why does the 3-button reset exist.

When I was making changes to the cplds, at one point I ended up with a cpld that caused the software to hangup so I couldn't reprogram it without cutting the links and using an external programmer but before I did that I tried adding the 3 button reset and it worked. (This might also work around the hangup you got programming the Atom cpld into the other boards but I've not tried it)
Again this might also be useful if a programming failure led to a half working cpld.

> Why if the CPLD version was valid on a 3-button reset, is it ignored?
 
An oversight at the time, that could be changed now.

>Writing the CPLD Type to the SD Card Volume ID. This actually makes me very nervous, because 
I move cards around all the time.

So do I, but I don't see it as a problem, as mentioned as soon as you move the SD to a new system and reset it, the volume name is changed to that system type. 
It only determines the cpld files offered on a blank system so if you move an SD with "3BIT" volume name to a six bit system without resetting it you will still get the correct cplds.

It's only an issue when moving between programmed and unprogrammed systems and even then it's better than before as it always defaulted to 6 bit but now you can explictly force the correct cpld choice by manually renaming the volume.

I'll look at the couple of suggestions above and I've also made a couple of other tweaks:
I enabled the GPIO pullups on the switches.  
MUX is enabled during vsync detection to keep the clamp pulses working (setting version to 0 to do vsync detection stops clamp pulses from being output on the RGB cpld but MUX overrides that)
If the YUV cpld was changed to be the same as the RGB cpld for the above P19 detection it would need that as well.
I tried adding those safety changes but it won't fit, perhaps you could look at it:
[YUV-CPLD.zip](https://github.com/hoglet67/RGBtoHDMI/files/4171506/YUV-CPLD.zip)

hoglet67:
Ian,

I've just done some testing of this on a 6-bit board, linking STB and P19 with a resistor.

The first thing the user is presented with is a message saying: CPLD is unprogrammed, which is not the case. We definitely need to give a different message in this case.

I then went ahead and reprogrammed the 3-bit firmware (in the 6-bit board) and that worked fine.

The problem was then getting back, because the 3-bit board marker has been written to the SD Card. The only way to revert from this situation is to reformat the SD Card, then do the three-button reset on the very first boot, which defaults to 6-bit firmware.

I didn't try this, but I think if anyone builds a fresh 3-bit board, with the 4 programming jumper wires present, they will need to jump through hoops to get the right choice of firmware initially (i.e. manually format a SD card and set the volume label appropriately). 

I think we are trying to be too clever here.

What I would suggest is we do the following:

1. Keep the code that detects that the 6-bit CPLD has been programmed on to a 3-bit board, with the following minor changes:
a) tighten up the test: look for a  specific change from DESIGN_RGB_TTL (GPIO22=0) to DESIGN_RGB_ANALOG (GPIO22=1). Only in this case enter force-reprogram mode.
b) only ever set GPIO22 as an output if DESIGN_RGB_ANALOG is read back in both cases.

What's compelling about this behaviour is the GPIO22 only remains an output if we are really sure there is an analog board connected.

2. Revert the code that writes a board marker to the SD card.

3. When force-reprogram mode is entered (for whatever reason), make the user explicitely choose the board type (ATOM / 3BIT / 6BIT), with a huge warning to seek help if they are not sure.

In addition, the status line should explain why force-reprogram mode is entered:
- CPLD Recovery Mode entered because CPLD is unprogrammed
- CPLD Recovery Mode entered because CPLD type (xxx) is unknown
- CPLD Recovery Mode entered because CPLD type (xxx) is incompatible with 3-bit board
- CPLD Recovery Mode entered because of 3-button reset

Then on the second screen with the list of CPLD firmware, the status line should indicate the board type selected by the user, just to give the user a chance to think a second time,

4. In the future, if we can get JTAG boundary scan working, then we can replace (3) with an automatic test.

Happy to help with this over the weekend.

Dave

IanSB:
Just thought of a way to protect against YUV in a 3 bit board:
During recovery mode only offer the RGB CPLD which is safe to get wrong.
Then the only way to program the YUV CPLD would be from a working six bit system.

hoglet67:
> Just thought of a way to protect against YUV in a 3 bit board:
> During recovery mode only offer the RGB CPLD which is safe to get wrong.
> Then the only way to program the YUV CPLD would be from a working six bit system.

That works I think.

I'm busy this evening, but have plenty of time over the weekend to continue this.

Dave

IanSB:
Thats' done.

There are now four recovery modes which are titled appropriately:
BLANK
UNKNOWN
WRONG
MANUAL

In any of the above modes, a new "recovery" folder in the cpld_firmware folder is used for programming and this contains copies of the 3bit BBC file and the 6 bit RGB file slightly renamed with "3bit" & "6bit".
This means it should normally be impossible to update the 3 bit board with the YUV firmware unless you start moving files in the firmware folder.
As far as I can tell this should be reasonably failsafe. If you program 6 bit into the 3 bit you get an error and can reprogram. If you program the 3 bit into the six bit it will mostly work in TTL mode but you have to use manual recovery (3 button reset) to reprogram the correct cpld.

hoglet67:
Just catching up.... This looks great, I'll try to merge tomorrow and create another release candidate.

IanSB:
I've not been able to get the pullups working so far, I'm using the sequence described on page 101 of the peripherals manual but it seems to have no effect:
https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2835/BCM2835-ARM-Peripherals.pdf

IanSB:
I built up a couple more 6 bit boards and noticed an issue with the RGB cpld and 24 Mhz mode 7
One board works fine but the other one has a messed up screen at some sampling offsets. Normal mode 7 works OK so I guess it's can't overclock as well as the other one but it's interesting that it only affects some offsets:
24 Mhz Offset 7:
capture0
24 Mhz Offset 6:
capture1
It affects all odd numbered sampling offsets but 7 is the worst.

If I program the V6.6 3 bit cpld into the problem board it works fine so the overclock is sensitive to the routing/complexity of the CPLD.

I suspect it's the swap bits MUX setting which has added to the complexity and that isn't really useful with the RGB cpld as it would give 4 levels of green in analog mode but still 3 levels of red / blue. 
Swap bits is potentially useful in the YUV CPLD as it gives 4 levels of Y and it might have been useful with the RGB cpld when used in mono mode with input on G only but as I've decided that all mono stuff belongs with the YUV CPLD it is probably redundant.

Do you think it's worth removing it to see if it fixes the overclock issue or can the CPLD be tweaked in some way to improve the routing?

hoglet67:
> Do you think it's worth removing it to see if it fixes the overclock issue or can the CPLD be tweaked in some way to improve the routing?

I've just tried, and it does seem to remove the issue.

I'm about to work on a couple of other bits, so I'll merge and push when I'm done.

Dave

PR #144, opened 2020-02-22
New pcb layouts + support for issue 3 analog board 

IanSB:
The new V3 pcb layout for the main "hat" now has 8 bit support (will require new CPLDs to support any 8 bit modes)
The extra 2 bits could be used for full 8 or 9 bit (using Vsync) TTL support or they could be used for additional comparator level detection in the existing 6 bit modes (like the swap bits option).

The new V3 analog board supports the AC/DC switch and has weak pull downs to prevent floating inputs. It also has an optional additional comparator on the underside for experimentation with enhanced U & V detection.
The DC mode seems to have eliminated the colour fringing I noticed on the Amstrad when calibrating.

I changed the terminate setting to "Input coupling" with options: 
DC/ High impedance
AC/ High impedance
DC/75R Termination
AC/75R Termination
This was to save a menu line as these menus are getting rather long.

The profiles have been updated to support this new option.
I found that AC mode was needed with Atom and it also helped reduce the mains hum with the dragon but most others worked fine in DC mode.
Note when setting up a profile, set it up first in DC mode, then switch to AC mode and adjust the clamp level to get the best results when using the DC values for the other settings. That allows users to switch between AC and DC mode without having to do major adjustments on the other values.

Your existing issue 2 analog boards should still work provided you select them in the interface setting on the front page (once for RGB and once for YUV) but this is untested.


hoglet67:
Hi Ian,

I'll take a look at the new V3 design over the weekend (I'm at the SW RISCOS show today).

BTW, Did you intend this pull request to be generated against master?

Dave

IanSB:
> BTW, Did you intend this pull request to be generated against master?

No, not sure how that changed from dev, I've fixed it now.

As far as I can tell, there are four CPLD inputs that aren't being used:
SW1, SW2, SW3 & LED1
These are all connected to GPIO pins and I assume that was for flexibility when designing the original.
I've taken the CPLD lines formerly connected to SW1 & 2 to the new 2 header pins and connected the other two to SW2 & SW3 via solder bridges (to avoid them floating). Cutting these bridges would free up those two inputs for expermentation and provide solder pads to connect the wires. 

However I just noticed that SW1 is mentioned in the RGB CPLD code:
process(sp_clk, SW1)
There isn't any other reference, was that something left over from when switches were processed in the CPLD?

hoglet67:
> There isn't any other reference, was that something left over from when switches were processed in the CPLD?
Yes, I think that's just a remnant.

IanSB:
Since we removed the swap bits option in the RGB cpld, the input mux setting and it's associated Vsync DAC setting had no real function but I have now made the input mux setting use Vsync as the sync source instead of Csync. This could be used for sync on green support and as it is less complex than the swap bits function it doesn't seem to affect the overclocking like that did.

Previously this wouldn't work as the signal was behind the clamp but now that DC coupling is available it should work. (I also cleaned up the SW1 reference)

Not sure if the 3 bit version needs rebuilding.

Note when you select this option in a BBC profile it starts rapidly switching in and out of mode 7 which shouldn't happen so I will investigate that as I think there is a long standing bug to do with loss of sync detection. (If you manually create a profile with input mux set and restart with it you don't get the rapid switching, just unsynced video but it then starts rapidly switching when you turn it off instead)

IanSB:
I was right that there was a long standing bug with loss of sync detection. I first noticed it when adjusting sync threshold levels with the analog board when the sync threshold was set wrong and producing noise but it was difficult to reproduce on demand.

In the above case with the new v73 CPLD, switching mux on caused sync to be sourced from the Vsync input which is not connected but pulled low by a 4.7K resistor. 
However, even with that pulldown there was enough crosstalk to put 2V spikes on that unconnected input and that was being interpreted as valid syncs which stopped the sync loss detection from working. 

I modified the sync code above to do some sanity checking so that if a vsync pulse is less than 64us or a hsync pulse is less than 2us, that is seen as loss of sync. I also added some hysteresis so that the sync state has to be valid for several frames before the state is changed. This seems to have vastly improved valid sync detection. I've tested this with all my different systems and it doesn't seem to have broken anything.

hoglet67:
Hi Ian,

I've updated to the latest Pi Software, and the v7.3 CPLD, and I'm seeing some strange behaviour.

This is with the V2 analog board connected to my BBC Master. I've manually selected the Issue 2 interface.

I'm seeing some loss of sync with the "BBC Micro" profile, but the "BBC Micro 24MHz" profiles works fine.  The only difference between the Sampling section in profiles is the Mux setting, which is off in the "BBC Micro 24MHz" profile, and on in the "BBC Micro" profile.

Changing it to off in in both seems to fix this issue.

> Not sure if the 3 bit version needs rebuilding.

No it doesn't; the JEDEC file is the same.

Dave

IanSB:
> Edit: Looking at the profile, I think the mux setting is incorrectly set to on in the Sampling= line in the "BBC Micro" profile.

It was set by error on the analog profile and I only tested the BBC on the TTL profile which is set correctly. I've committed a fix

Issue #140, opened 2020-02-06
Release candidate


IanSB:
I've just done some testing with the PC and that seems to be behaving itself.

Do you want to make up a release candidate zip and I'll do a run through of all the machines again and if that looks good we can go for a release on the weekend.

hoglet67:
https://github.com/hoglet67/RGBtoHDMI/releases/tag/20200206_53bd391

hoglet67:
My only thought was whether we wanted to prune CPLD versions that are included. I'm in two minds about this, as it's incredibly useful to us to be able to go back and forth.

Plus it's further work....

I fully expect we'll see issues, and need to make other changes over the next few weeks.

hoglet67:
One thing I need to check before we announce is that .jed files in the various vhdl directories reflect the latest version. I'm pretty sure some of them are stale.

I'd also like to delete the .xsvf files in the vhdl directories, as these duplicate the ones in the src/scripts/cpld_firmware directory.

Dave

IanSB:
> My only thought was whether we wanted to prune CPLD versions that are included. I'm in two minds about this, as it's incredibly useful to us to be able to go back and forth.

I think that might be confusing for new users especially when they are expected to switch between the YUV and RGB cplds on a semi regular basis.

How about putting them in "old" sub-folders in each cpld folder. They wouldn't show up normally in the menus but could be moved out into the main folder if required or you could make it so they showed up by enumerating the "old" sub-folders if debug was enabled in the settings menu.

BTW spoke too soon about the PC, I've just fixed a sync bug that was causing auto switch to get stuck sometimes. I'll send a pull request soon.

IanSB:
I've implemented the "old" folder for cplds when debug enabled.

hoglet67:
New release candidate:
https://github.com/hoglet67/RGBtoHDMI/releases/tag/20200208_dda4935

This morning I'm really just made 2 changes:
- create a RGB CPLD v72 that removes the swap bits functionality as suggested
- in CPLD Recovery Mode, automatically bring up the menu, and add further help/warnings

IanSB:
A few comments so far:

All the txt files (Profiles & Config.txt etc) are in unix LF format. Did you change that deliberately or was it done via Github? (I get warnings about eol conversions on my windows Github client)
Realistically, most users will be using windows to look at/ edit the txt files so maybe it is better to release in Windows CR/LF format so they can be edited with notepad.

[...]

PR #145, opened 2020-02-28
Four new pcb designs

IanSB:
Four new pcb designs:
1. V3 6/8 bit board with 12 way connector (Standard version).
2. V3 Template version. Same as above plus tabs and alignment holes.
3. V3 analog board, now with 6 way IDC header support.
4. V1 TTL buffer board. This is based on the analog board and uses a 74LS08 to buffer RGB & sync like the 3 bit Electron buffer.
Not really sure how useful this will be but it was an easy mod of the analog board.
It can also be configured as a "straight through" 6 way IDC to 10/12 way IDC converter using the jumpers which could be useful for testing those TTL computers like the beeb that work both with the analog board and directly with the TTL input. With this board, both analog and TTL versions can use the same cables.

Can you do a sanity check on all the above as I will be uploading to JLC soon.

BTW I've not found any issues with the latest release candidate so far.


hoglet67:
H Ian,

I've have a quick look at each of the boards and found a couple of things you might want to check.

1. If you do a DRC on the analog and TTL frontends, it reports the zone fills are out of date. This means that changes have been made since the zones fills were last updated. It's probably worth doing one final DRC on each of these (to update the fills), then regenerate the gerber files.

2. The TTL frontend is missing a decoupling capacitor for the 74LS08

3. The protruding PCB spigot for the 6-pin IDC connector looks like it will extend ~0.3mm beyond the outer edge of the plastic case. That's different to your mockup here, where it looks like it's internal.
https://github.com/hoglet67/RGBtoHDMI/issues/140#issuecomment-591239075
It's possible I've mis-measured, but could you double check this.

Dave

IanSB:
> zone fills are out of date

I made some last minute changes to the silk screen but forgot to refill.

> The TTL frontend is missing a decoupling capacitor for the 74LS08

Good catch I missed that!

> looks like it will extend ~0.3mm beyond the outer edge of the plastic case

That was by design.
I mocked up my original idea with it internal to the case but the problem is that you have to make a very accurate notch in the case to fit around the body of the 6 way IDC header and there are too many variables in the system that affect the final position of the 6 way header which also means boards will not be interchangeable.
(i.e. play in the 40 way and 4 way headers, angle of the 4 way headers, mounting screw holes, solder position of the 6 way IDC on the solder tabs etc)

Extending the 'spigot' to the outside edge of the case means the case notch only needs to accomodate the pcb and the soldered pins of the IDC and the body of the IDC hides the case notch so it doesn't have to be very accurate and can be slightly oversized to accomodate all the play in the system.
However, after re-measuring and printing out the layout at 1:1 for a mockup I think it probably sticks out a bit too much now so I'll reduce it by 0.2mm or so.

hoglet67:
> That was by design

But doesn't that mean if you want to use only the Picoblade connector, you will be forced to cut a slot in the case, and have a bit of PCB sticking out as well?

IanSB:
> you will be forced to cut a slot in the case, and have a bit of PCB sticking out as well

Or hacksaw off part or all of the 'spigot'.

It's always possible to do an alternative layout with a shorter spigot but I think the longer one is the best for experimentation at the moment as you can always cut or file a bit off but it's harder to add it on.

I've updated the boards with the suggested changes and reduced the length of the spigot by ~0.5mm as it was too long.

PR #146, opened 2020-03-02
Changes to help setting up new geometry settings

IanSB:
Adjusting the pixel aspect ratio now shows helper info with the actual current aspect ratio.
Also the pixel aspect ratio is shown on the Source Summary info page.

There is now a "Fine Set Clock" option in the geometry Setup Mode. Selecting this changes the clock increment from +-1000 to +-1. 
With the acceleration you get from holding the up/down buttons it should now be possible to accurately set any clock frequency relatively quickly. 

Also when you adjust the clock or line length, the setup mode is automatically forced to "Set Clock/Line" if in non-clock adjust modes. This is because it is only useful to adjust these values in the setup mode which disables the clock tolerance setting. If you adjusted with clock tolerance enabled you get very misleading results.

Note Offset/Min/Max adjustments don't force the setup modes as it can sometimes be useful to adjust those without being in the setup modes.

Also I added a pulldown resistor to the 6/8 bit boards on the Data line (GPIO0). This line is weakly pulled high on other boards so it should be possible to auto detect the 8 bit board by reading this bit on bootup.


IanSB:
I just identified a hazard which I think I've fixed:

When the menus got longer, even the 8x8 font took too long to write to the screen to keep up 50Hz refresh when menus were displayed so I decided to abandon 50Hz refresh and assume 25Hz refresh during menus which I why I switched to the larger teletext font for all modes.
This meant that the OSD was being updated just before, during and just after field sync causing every other field sync to be skipped and thus 25 Hz refresh.

However I just noticed an issue when looking at the empty "Calibration Detail" menu which started glitching, something I hadn't seen before. 
As this was a very short menu, it turned out that  osd_update_fast was completing in the middle of field sync which caused the field timing measurement to get variable results and that caused continuous timing recalibration. I think this would vary depending in the actual BBC clock frequency and PPM error so may not be easily reproducible.

To avoid this hazard, I put a wait for field sync before osd_update_fast.
This guarantees only 25Hz refesh with the OSD on but as this was already the case in most circumstances it doesn't really change anything.
However this does mean OSD update finishes slightly later than before so we should check for any visual consequences of that.

PR #147, opened 2020-03-03
Improve sync detection code so helper info works properly

IanSB:
Previously the sync helper info in the geometry menu just showed the current sync setting unless you were in autoswitch PC mode. Now it shows the detected sync type irrespective of the current sync setting.

There is a limitation:
If you have selected one of the two composite sync types (composite or inverted composite) it will display what it thinks is the correct one of those two.
If you have selected one of the four separate sync types (+H+V +H-V -H+V -H-V) it will display what it thinks is the correct one of those four.

It will not try to identify if the sync is composite when you have selected separate or separate when you have selected composite.

This is due to two issues:
1. In TTL mode there can be crosstalk on the unused vsync input causing misleading readings
2. The analog interface connects the vsync input to the Green/Y input via a comparator so there will be misleading pulses on there depending on the source and the comparator setting.


hoglet67:
Ian,

Do you want me to do a new new release candidate this includes this?

Or shall I wait a day or to in case there is more to follow?

I'm a bit distracted on other matters at the moment, do haven't yet had a change to look at the Wiki.

Dave

hoglet67:
Ian,

Sorry, I missed your earlier message about holding off on the RC.

I've deleted that release candidate and I'll remake it when you are ready.

Dave

IanSB:
> Do you want me to do a new new release candidate this includes this?

Yes this can be included, I've just finished testing.

Most of the recent changes have been driven by writing the documentation and discovering these issues during the process.

IanSB:
I'm seeing a possible issue with sync timing so hold off on the next RC for now.

IanSB:
I'd been noticing a problem where occasionally during a genlock resync I would get an interlaced changed error. It always seemed to happen after 10 resyncs but I couldn't work out the significance of the number 10 although I thought the problem was indicative of a frame being dropped.

It turned out the cause was the logging of the resync message itself:
The "INFO: ReSync: xx" and "INFO: Locked" log messages were transmitted during the frame capture loop so had to take minimal time and from previous results I had assumed the transmit FIFO buffer was large enough to accomodate those messages but after looking at the documentation, the FIFO is only 8 bytes which meant the logging was stalling the frame capture loop. It takes about 1.35 video lines to send each char at 115200 baud so it was taking multiple video lines to send the message between the end of capture and the start of vertical sync.
It just so happened that when the ReSync counter hit 10, printing that 1 extra char (which took an extra 1.35 lines) was enough to cause the frame drop!
At the time the logging was introduced, fewer lines were captured per frame so it wasn't noticeable as there was extra time but setting the max capture to 288 meant that more lines were captured on a 1600x1200 monitor than a 1080p monitor (only 270) so I think it only happened at that resolution.

I've reworked the logging to drop the "INFO:" prefix (which will speed up logging in general) and then reduced the genlock logging to 6 chars + CR/LF so it should always fit in the FIFO and that seems to have fixed things. (The ReSync counter will now wrap at 99)

I also found a better way to filter noise in the sync source by further deglitching CSYNC in software so have reverted my previous attempts.

I think this can be made into a new release candidate if you don't find any obvious issues with it.

hoglet67:
> I think this can be made into a new release candidate if you don't find any obvious issues with it.

I've uploaded a new release 

Not actually tested it yet, as I'm burried in other things at the moment. I've got a Spec Next on load for a few days, and I'd really like to get HDMI working with the Beeb core.

PR #148, opened 2020-03-17
Low resolution bug fixes 

IanSB:
I found quite a few bugs when testing the low resolutions being used by Aotta:

1. After selecting 1024x600@50Hz (first resolution in sorted list) you could no longer change the scaling setting.
2. 800x480 resolution added as there are lots of lo-res mini monitors on ebay using this size.
3. Inhibit double width or double height if the resulting size exceeded the screen size.
4. Setting 640x480@50Hz and 800x480@50Hz resulted in the read back of the horizontal screen size from the registers being doubled so it incorrectly assumed a screen size of 1280x480 or 1600x480. I think this meant that the HDMI clock rate was too low so the Pi was using a double clock rate. There is probably a register bit somewhere that indicates this but for the time being tests are made for those resolutions and they are adjusted accordingly.
5. The overscan calculation was badly wrong with some combinations of screen resolution and profile e.g. 1280x800 with Spectrum profile.
6. The cpld programming code was adjusted to keep the clock on average below the logic 1 threshold and some timings were increased.


hoglet67:
Merged, and a new release created.

PR #150, opened 2020-04-14
Fix HDMI not working on bootup with some SD cards

IanSB:
Fix HDMI not working on bootup with some SD cards by detecting invalid HDMI resolution and rebooting.

Also update 6 bit and 8bit PCB layouts

As mentioned, can you run a calibration on the default BBC micro profile in TTL and Analog RGB modes (not legacy) before making another RC


hoglet67:
I've just tried running the calibration on each of my Model B's and they end up a bit different:

Beeb 1 (Issue 7):
- Mode 0-6 = 2 (delay 5) (errors at 0)
- Mode 7 = 3 (delay 1)

Beeb 2 (Issue 7):
- Mode 0-6 = 4 (delay 5)  (errors at 1,2)
- Mode 7 = 6 (delay 1)

Beeb 3 (Issue 4):
- Mode 0-6 = 0 (delay 3)  (errors at 1,2)
- Mode 7 = 6 (delay 1)

I get the similar results with the V2 Analog Board.

So I've tweaked the profiles to:
- Mode 0-6 = 4 (delay 5)
- Mode 7 = 6 (delay 1)

But people will need to recalibrate!

Dave

PR #151, opened 2020-04-27
More bug fixes and prodest profile

IanSB:
Fix mono palettes so vsync indicator bar works
Fix clipping error on chars per line which was still putting a black line on the right hand edge with certain resolution and profile combinations. e.g. zx81 at 1024x768
Put the CPLD clock (6 or 8 times pixel clock) in source summary (useful to see if CPLD is being overclocked)


hoglet67:
Ian,

Just let me know if you want me to push another build...

Maybe we need to stop calling the builds release candidates, and just call the incremental releases.

Dave

IanSB:
Yes, I was thinking the same. I've tried to avoid making any major changes recently and only done bug fixes and tweaks for issues raised on the forum like the screencap size.
A lot of user functionality issues have been ironed out by writing the documentation which is now mostly complete so I don't think there will be many more changes due to that.

Hold off until Sunday in case aotta tries the Spectrum profile and finds some problem with that.
If all looks OK by then, make this an official release and merge it to the master branch.

There is still work to be done like the in band palette control and NTSC artifact colours together with 8 bit support so this would be a "0.90" style release and further work will continue in dev.

BTW Have you tried the new analog board?

hoglet67:
> BTW Have you tried the new analog board?

No I haven't - in fact I haven't even ordered the bits yet.

I've mostly been working on other projects.

IanSB:
OK I think it's time to do a formal master release.

hoglet67:
I'll merge back to master and do a release from there tomorrow.

hoglet67:
Ian,

I've merged back to master and uploaded a release.

I've marked it as a stable release (i.e. not a pre-release) but I'm fully expecting in the comings days and weeks that updates will continue to happen on master, and further releases can happen as-and-when needed.

I suggest we simply refer to this as the "20200504" release.

I've also added a some missing images and PDF schematics to the Wiki. I used the photo you posted on Stardot of the V3 analog boards as I still haven't build mine up yet.

Thanks for all the work you have done on this. It's nice to see quite a few people starting to experiment with these.

Dave

PR #157, opened 2020-09-15
Bug fixes and improvements to new 12 bit and Atari 800XL support

IanSB:
This is mainly bug fixes and minor tweaks to the new features but it does fix a nasty crash bug on screencaps so I think I new release candidate should be made.


hoglet67:
Ian,

The two analog boards I have both say "MAX5259 DAC Issue 2 (c)2019 IanB" on the PCB, so I assume these are issue 2.

It appears these have been dropped  ;-(

Is there a good reason for this? I'l prefer not to have to ditch these.

Dave

IanSB:
I haven't actually removed support, just hidden it. There are a lot of analog boards going out at the moment and I was concerned that if end users started messing around with that setting their boards would stop working.
Support can be re-enabled by tweaking the code as suggested in the comments here
https://github.com/hoglet67/RGBtoHDMI/pull/157/commits/531f6351cb6824c3f12a2fa4362ce9c4e9af8232

However I could add it as an option in default.txt to re-enable support if you would prefer that
Something like old_analog_support=0 / 1

hoglet67:
> However I could add it as an option in default.txt to re-enable support if you would prefer that
> Something like old_analog_support=0 / 1

That work work for me - I'd just prefer to avoid having to make code changes as part of a test cycle.

Am I the only person still using a V1 or V2 analog board?

IanSB:
>Am I the only person still using a V1 or V2 analog board?

Yes, I rebuilt mine to V3

Actually had a better idea:
The pre-built issue 4 boards can be auto detected so if one of those is fitted it disables selection and if not, selection still works as before.

IanSB:
Ok that's done:
If issue 4 analog board detected, no selection is possible.
If the above not detected then analog board selection of issues1-3 is possible as before

hoglet67:
Thanks Ian, that's a much better solution.

There's a new release build here:
https://github.com/hoglet67/RGBtoHDMI/releases/tag/20200908_ff298ac

PR #158, opened 2020-09-30
Auto resolution / refresh, log saver and many bug-fixes

IanSB:
Auto resolution and refresh detection
There are 3 new resolutions:
Default@EDID - Selects the resolution and refresh based on the EDID and source computer refresh.
Will only select 50Hz refresh if the EDID indicates support. This is likely to not work with most old DVI monitors except HP ones and it will probably work with more modern HDMI monitors.
Auto@50Hz-60Hz - Always outputs 50Hz if the source is 50Hz. This will result in a blank screen or out of range message on some monitors. Never goes above 60Hz
Auto@Unlimited - Always tracks the input refresh rate even if it's greater than 60Hz
All refresh changing is done by slowing or speeding the HDMI clock.

Default@EDID is the new default with the CPLD version and Auto@50Hz-60Hz is the default with the simple board.

Menu-power-on recovery has changed: Now it reselects the above defaults as the resolution but also completely disables 50Hz support until the next reset (A message is displayed to indicate that)
This is to recover from any monitor that won't work at 50Hz despite it's EDID indicating it would work.

A message is displayed on startup indicating the current resolution and refresh rate.

The serial port log is now written to memory as well and can be saved from the info menu together with the current monitor's EDID info to help in debugging problems.

I've added divider support back into the common (24Mhz) CPLD by removing sync inversion and restricted the profiles used by it to just the beeb/master/electron ones.

There have also been many bug fixes following testing with most supported computers.


IanSB:
The more recent commits have reworked the CPLDs and split BBC support from the main RGB CPLD
In 6-12 bit mode there are now 3 CPLDs:

6-12_BIT_BBC_CPLD_v77
6-12_BIT_RGB_CPLD_v90
6-12_BIT_YUV_CPLD_v90


6-12_BIT_BBC_CPLD_v77:
The BBC CPLD is functionally similar to the 3bit v66 CPLD and supports 24 Mhz mode 7 but in addition supports the analog interface and supports working on the 12 bit board by adding the divider bit to the register chain. Two options have had to be removed to get that to fit: The mux bit used for selecting sync on green (Vsync as sync source) and the sync invert bit used for separate H&V sync sources. As a result this CPLD no longer supports profiles that use those and is effectively limited to the same profiles as the 3 bit board and it shares the same profile folder which has been renamed to "3-12_BIT_BBC"

6-12_BIT_RGB_CPLD_v90:
The RGB CPLD has had BBC mode 7 support removed (although you could still create a BBC profile if you had a symmetrical 12 Mhz clock). In it's place the following are supported:
Sync Inversion
Sync on Green
Separate syncs
3 and 4 dividers as well as 6 and 8 (Which means higher pixel clocks without overclocking the CPLD)
4 level analog RGB
9 and 12 bits per pixel capture
1 Bit per pixel capture (Can be used with very high pixel clocks to pass 8 pixels per psync edge instead of 4, 2 or 1
3 bit delay (Required for 1bpp mode as 8 bits per psync)

I have tested this at 48 Mhz pixel clock with divider set to 3 and that seemed to work OK (CPLD = 144Mhz). In theory with some CPLDs it might work up to a 64 Mhz clock (CPLD = 192Mhz)
This supports all existing RGB profiles except the BBC ones.

6-12_BIT_YUV_CPLD_v90:
This is unchanged from previous pull requests

I have also re-checked in the code for the v85 CPLD (as "vhdl_RGB_12bit_old_V8") which is half way between the above v77 and v90 in that it supports BBC mode 7 (but not 24Mhz mode 7) and doesn't support  3 & 4 dividers or 1bpp mode which are used in the Atari ST mono mode profile and should also be used by an Archimedes mono mode profile. I haven't included it in the main CPLD folder although it is available in the old folder and should still work but it would require another set of profiles with 24Mhz mode 7 and Atari ST support removed.

I've also added enforced CPLD upgrade on a fresh install as testing is getting quite complex so only limited testing is done on old CPLDs (this only works on 6-12 bit boards). After a clean install, if the CPLD versions are not the current ones it will request an update using the recovery menu but will only list the 6-12 bit CPLDs. This only happens on a clean install so you can downgrade again afterwards and it won't complain until the next clean install.

Some of the sampling menu options have beem renamed to conform to the terminology used in LCD VGA interfaces which use phase and clock:
Offsets -> Sample phase
Divider -> Clock Multiplier (options x3 x4 x6 x8)
Delay -> Pixel H Offset (To match H offset in the geometry menu)

IanSB:
I've made some improvements to 24 Mhz mode 7 auto calibration:
I noticed that on some cpld boards it would get the calibration completely wrong when using x8 multiplier but seem to work OK when using x6 multiplier. After some investigation it turned out that the "half" setting was not symmetrical around half a pixel when the CPLD was very overclocked so the code that determined the half setting and then swapped the error tables had erroneous results.

Now if auto calibration is run in 24 Mhz mode it separately calibrates all four combinations of half and 24Mhz odd/even and picks the one with the lowest errors and that seems to work OK. 
This also means there only needs to be one 24 Mhz profile and it now defaults to using the x6 clock multiplier as that only overclocks the CPLD to 144 MHz which is much more likely to work and still better than the standard mode 7 but the user can also try x8 as a last resort. 

Not sure if it is better to use x8 as the default and let the user manually reduce to x6 if it doesn't work  or should we include 2 profiles:
BBC_6x24Mhz_Mode_7
BBC_8x24Mhz_Mode_7

Please check that normal calibration still works as expected.

hoglet67:
Hi Ian,

I should have time over the next couple of days to try this out.

Assuming everything seems fine, shall I merge everything in the pull request?

Dave

IanSB:
>Assuming everything seems fine, shall I merge everything in the pull request?

Yes, but please check before making a pre-release build in case any last minute fixes are in the pipeline.

I'd like to get this released this week if possible so we can get some feedback on the new EDID reading code and if it is likely to cause any problems. If things look good with that I'd like to do a formal release version soon.

hoglet67:
Hi Ian,

Sorry for the delay in merging this, I got rather snowed in by other work, and some domestic stuff.

So far I've just tested in on:
- an Atom with the Atom CPLD
- a Master with a 6-bit Issue 2 board

On the Master, I was immediately thrown into the CPLD Recovery menu, and I selected the BBC CPLD v77. I assume this is expected.

The EDID stuff seems to be working great. It detectected my 1920x1080 TV and my 1600x1200 HP LCD. In both cases, a refresh rate appropriate to the source was picked. This is a really nice feature.

I also tried the new BBC 24MHz calibration code, and that gets a perfect result. I can try some other machines later.

Just ping me when you are ready for a pre-release build.

Also, if there are specific hardware combinations you want me to try, then please just ask.

Dave

P.S. And the Postman has just delivered a package, thanks for these.

PR #159, opened 2020-10-08
Fix some problems with continuous calibration

IanSB:
I found some issues with continuous calibration under certain extreme conditions and this commit addresses those problems.    

>On the Master, I was immediately thrown into the CPLD Recovery menu, and I selected the BBC CPLD v77. I assume this is expected.

Yes, but you should have only been offered 6-12bit CPLD versions unlike a new install which offers 3 bit ones as well with all the warning text.

>Also, if there are specific hardware combinations you want me to try, then please just ask.

I'm currently testing 16 boards simultaneously, all connected to different vintage computers  but there are a few more things you can test if possible but please include the above commit before any further testing:

Leave things running for an hour or two and then check the logs.(You can use the new save log feature so don't need to connect to the serial port)
Ideally you should see an occasional update of the PLL or genlock like this:
*PLL
Locked
Sync01
Locked
Sync02
Locked
and hopefully no dropout of capture and loop around (also worth turning on the vsync indicator to check that is stable)

When loading logs, note they will all have the same datestamp which can confuse some text editors if you already have another log loaded so make sure any open log files are closed before looking at others.

My Electron has a huge number of entries for PLL and genlock which either means it's clock is unstable or there is something about it's sync output that is confusing the continuous calibration. It has so many entries that I ended up dynamically increasing the PLL recalibrate window from 1 PPM up to 6 PPM in steps if it sees lots of entries so you may see *PLL2 *PLL3 etc which means the window has increased to 2 PPM, 3 PPM etc.
Can you check with your Electron and see what sort of log entries you get

Can you also check with your Superboard as there is now de-jitter code to fix the vertical jitter caused by RC monostable V sync pulses. The profiles may need tweaking as there have been many changes since they were created.

I recall you mentioned previously testing with a CPC so can you check the analog interface on that and your Spectrum and Atom (I included an Atom lead and the beeb lead works with the CPC via the analog interface due to the protection diode on +5v).

Also test the beeb though the analog interface especially mode 7 to see how much impact on calibration that has (asymmetric clocks can exceed the bandwidth of the analog interface worst case I think).


hoglet67:
Ian,

I've merged this, and done some testing as I've been doing other bits and pieces. No problems so far with the Atom, Electron or BBC. I've been checking the log file, and it looks very much like your example above.

In terms of other machines, the CPC has gone back to it's owner, so I won't be able to test that.

I'll dig out the Superboard over the weekend.

There was one other small change I'd like you to consider. Could we rename the "Master 128" profile to "BBC Master". I share a device between my BBC Micro and my BBC Master, and so I'm constantlt changing between these profiles, and they ate currently a long way apart in the menu,

Dave

PR #160, opened 2020-10-09
Rename some BBC profiles

IanSB:
>No problems so far with the Atom, Electron or BBC. I've been checking the log file, and it looks very much like your example above

I guess that means my Electron's clock is very unstable.

>Could we rename the "Master 128" profile to "BBC Master"

I've renamed the above and also renamed Atom and Electron to Acorn Atom and Acorn Electron as that puts the Electron Closer to the BBC profiles

I think you can make a build of this version if the above looks OK to you as I haven't found any other issues so far.

Depending on any feedback, one possible additional feature that occurred to me would be to add a text file with a list of monitor IDs that don't work with the EDID to use as an exceptions list. i.e. it would contain monitors that do work at 50Hz even though their EDID doesn't indicate that and monitors that don't work at 50Hz when their EDID indicates they do.


hoglet67:
Here's a build of this version:
https://github.com/hoglet67/RGBtoHDMI/releases/tag/20201010_4c8ca86

PR #163, opened 2020-11-11
Update CPLD moved to main menu

IanSB:
Move Update CPLD  to main menu plus other minor tweaks

Please make an interim release build with all recent changes for testing.
Also I have updated the wiki quick start guide and reference guide to reflect all recent changes so perhaps you could check that


hoglet67:
There is a new dev release here:
https://github.com/hoglet67/RGBtoHDMI/releases/tag/20201112_49258e2

Quickly tested on Atom (using original adapter) and BBC Model B.

PR #164, opened 2020-11-13
Fix issue with clock tolerance compensation code

IanSB:
Unfortunately I found a 'showstopper' bug while testing:
I made some changes to the clock tolerance code to try to cope with higher multipliers but that resulted in a condition where if sync was intermittent (which can easily happen when changing profiles in analog mode) the CPLD clock would rapidly increase to it's cap of 200Mhz and on some boards that would result in a hard lockup due to loss of psync.

Could you replace the last build with this updated one


hoglet67:
No worries, here's an updated release:
https://github.com/hoglet67/RGBtoHDMI/releases/tag/20201113_0471da6

I've removed the old one from yesterday.

PR #166, opened 2020-11-24
Improved mode 7 auto calibration, NTSC profiles & other tweaks

IanSB:
I've added the auto multiplier ranging as discussed. There is a new "Multiplier Range" setting in the sampling menu which only shows up in mode 7 on the BBC CPLD with options Auto or Fixed

Auto: Starts auto calibration with multiplier set at x8. If more than 0 errors then it increases to x12 and runs calibration again.
Fixed: Leaves multiplier unchanged

If the multiplier is set to x16 then then the Auto setting is ignored and it just calibrates at x16 (This isn't selected automatically as it doesn't work reliably on all cplds)

I realised why the half setting was behaving strangely in x12 and x16 mode: It actually introduces a quarter pixel shift in those modes so you get invalid values when you swap the metrics by half a pixel.

I noticed an issue with odd/even which has been present since it was introduced: There was a pixel shift when switching from odd to even and that was now was quite noticeable when adjusting the sampling phase (offset) setting which happened as you crossed the boundary (e.g. from 5 to 6 in x12 mode). I've fixed that so there are new BBC CPLDs V68 (3bit) & V79 (6-12bit).

I've added a lot of NTSC profiles and fixed others like Commodore 128, some I've been able to test but others are untested. They make use of the sub-profile auto switching system originally created for PC use so you just select the appropriate computer and it auto selects the 50Hz or 60Hz profile depending on the source timing. (You can actually switch ZX80 and ZX81 from 50Hz to 60Hz while they are running by holding a diode or resistor respectively in place and they work OK. 

Can you review the calibration system as I've made quite a lot of changes there recently including the previous pull request including adding 12bpp support which seems quite slow (Can be tested with the Nula profiles which work but give 50% brightness due to only the top bit being used)

I also changed 1280x1024 to interpolated scaling as discussed.

If this looks OK, please make a new interim build and I will do a full test on that.


hoglet67:
Hi Ian,

Sorry for the delay, I had a bit of an off-day yesterday.

I've just given this a try, on a Model B with C48 tweaked to get an "interesting" number of errors (i.e. correctable, with some effort).

I'm finding the statistics a bit inconsistent, suggesting a bug somewhere.

The simplest way to see this is to switch to Fixed/12x and calibrate. As it does the calibration, there are places where it sees 0 errors, but on the final Calibration Summary screen there are a large number of errors shown:
capture0
capture1
capture2

The resultant screen looks perfect.

Here's the log:
[Log.txt](https://github.com/hoglet67/RGBtoHDMI/files/5597575/Log.txt)

It looks like what's happening is that the errors reported in the Calibration Summary screen comes from the last of the four runs, even though this is not the one that was selected.

As I said, I don't think this impacts the picture quality.

Dave

IanSB:
I missed out *errors = min_errors; when copying the minimum values. This wasn't obvious on my system because the errors on the final run were 0.
I also fixed a minor issue with the logging.
The above commit should fix those

hoglet67:
Thanks for that fix Ian. It's looking good to me.

I get significantly better results with the 12x multiplier, so that's a nice addition.

> If this looks OK, please make a new interim build and I will do a full test on that.

There's a new dev build here:
https://github.com/hoglet67/RGBtoHDMI/releases/tag/20201126_2d102ae

PR #167, opened 2020-12-01
Amiga adapter with support for SuperDenise

c0pperdragon:
After receiving some feedback of users with an actual SuperDenise chip, this adapter version can now be considered to work.


dRwH0:
Awesome work. Can you also list the BOM for this with your selections from Farnell or similar ?

hoglet67:
Just to say I've not forgotton this pull request.

I'll merge it back to master after we have merged the current development branch.

SirCathal:
I can confirm that it works fine with the normal and the super Denise Chip. 
Would be great if you could accept the pull request.

hoglet67:
Sorry for the delay.

PR #172, opened 2021-01-14
HDMI, resolution & refresh menu options, updated NTSC artifact & bugfixes

IanSB:
New main menu options:
HDMI mode to fix Jan Beta's capture card issue
Separate resolution and refresh options, now fixed and custom resolutions can also auto switch between 50 Hz and 60 Hz
Improved NTSC artifact code for Tandy CoCo 1 & 2 / dragon (could also work on Atom but not enabled in profile for now)
Sampling menu DAC options now display voltage levels and have "Disabled" option for unused DACs
Support for black scanlines in 12bpp mode (level not adjustable)
Also support for 50% scanlines in 12bpp mode but requires overclocking and will likely only work on Amiga.
Powerup resolution message can be disabled.
default_config.txt no longer needed, config.txt is modified instead of being re-created from default_config.txt. This means only config.txt has to be edited for any manual override settings.
Improved swap aspect option now works on Amiga
(Should work on a beeb too but not useful)

I found a possible Pi bug with nearest neighbour scaling:
When experimenting with 50% Amiga scanlines I found that I got glitches in 1920x1080 mode but not other modes like 1600x1200 or 1280x1024 so the extra screen DMA was stealing too many cycles.
To work around this I created a 1440x1080 mode which didn't have glitches and would still be displayed sharply on many 1080p monitors that had a 1:1 pixel mode.
However, many TVs don't have that option so I also created another mode: 1440(1920)x1080 which had a 1440x1080 display scaled from a 720x540 buffer with 240 pixel overscan on either side (which is similar to your original way of getting a screen output). This appears to work at first glance but there is something wrong with the integer scaling which makes the output look ragged.

e.g.
Can you try this with a beeb and see if you think it is Pi issue or a problem with the setup:
Put lots of text with diagonals on screen in mode 0 like //// and >>>> <<<< etc and select 1920x1080 with integer scaling. Then select 1440(1920)x1080 resolution and compare the above chars which will have steps in them.
According to the monitor it is receiving a 1920x1080 source in this mode and the frame buffer is 720x540 so it looks like it should work but the scaling coefficients must be miscalculated in some way.
If you change to integer soft scaling it looks the same at both resolutions.

Hold off on a release build for now.


hoglet67:
> Can you try this with a beeb and see if you think it is Pi issue or a problem with the setup:
> Put lots of text with diagonals on screen in mode 0 like //// and >>>> <<<< etc and select 1920x1080 with integer scaling. Then select 1440(1920)x1080 resolution and compare the above chars which will have steps in them.
> According to the monitor it is receiving a 1920x1080 source in this mode and the frame buffer is 720x540 so it looks like it should work but the scaling coefficients must be miscalculated in some way.
> If you change to integer soft scaling it looks the same at both resolutions.

Yes, I do see that effect.

This is the mode being used?
```
## 1440x1080 @ 50Hz
hdmi_group=2
hdmi_mode=87
hdmi_cvt=1440 1080 50 1 0 0 0
overscan_left=180
overscan_right=180
```
I'm confused how this is intended to work.

Aren't you then scaling a 720x540 frame buffer up to 1080 width? (1440-180-180)

I've also noticed some issues when changing the refresh rate in the menu. You are promted for an automatic reboot (expected), but unless you also manually save the setting, then the change is lost.

IanSB:
>Aren't you then scaling a 720x540 frame buffer up to 1080 width? (1440-180-180)

Yes, I misunderstood how the overscan worked, I was hoping it would scale up to 1440 then add on the padding but, as you say, it's subtracting from 1440 then scaling up instead so it's not going to work for integer scaling. It does actually work in interpolated mode because that looks the same irrespective of how the final output is achieved.

>I've also noticed some issues when changing the refresh rate in the menu. You are promted for an automatic reboot (expected), but unless you also manually save the setting, then the change is lost.

Not sure what you mean by that, the refresh rate is not saved by the save config option, it's written to the config.txt file as soon as you change it. Can you give me a sequence to reproduce that.

IanSB:
I fixed the scaling issue:
You need this resolution file:

1440(1920)x1080 @  60Hz
hdmi_group=2 
hdmi_mode=87  
hdmi_cvt=1920 1080 60 1 0 0 0  
overscan_left=240  
overscan_right=240  

This gives the same resolution readback from the Pi's video registers as a normal 1920x1080 mode so you must also subtract the overscan values from the resolution readback which gives 1440x1080 which then means a 720x540 buffer is created.

Everything then scales as expected

hoglet67:
Does that also fix the scanline issue?

The HDMI pixel clock will still be the same as in 1920x1080, so I would expect the peak memory bandwidth (during the active part of the line) to be the same.

Have you experimented with the reduced blanking modes. These use a lower HDMI pixel clock, so should reduce peak memory bandwidth. The blanking is less, which may cause problems though....

hoglet67:
> Can you give me a sequence to reproduce that.

OK, it's a bit more difficult that I first thought, but try this....

- Start with a blank SD card and copy the release build
- Boot the system (I used a Model B)
- Hold SW3 down to auto calibrate
- Enter the Menu, and Save the Configuration
- Then change refresh rate, e.g. to 50Hz
- You are promoted that a reboot will occur on menu exit
- Select Return to exit the menu
- Allow system to reboot

Refresh rate is still EDID 50-60Hz

I think the issue is triggered once you have a saved configuration.

At this point, the only way to change the refresh rate and get it to stick is to manually save the configuration before rebooting.

Dave

IanSB:
> Refresh rate is still EDID 50-60Hz

OK I fixed that.

IanSB:
> Does that also fix the scanline issue?

Yes it does but I'll look at reduced blanking as well

PR #180, opened 2021-01-31
Change from RGB565 to ARGB4444, Flash Floppy OSD & teletext calibration

IanSB:
The frame buffer has changed format from RGB565 to ARGB4444 (i.e. 12 bits RGB + 4 bits Alpha) by manipulating the display list as there didn't seem to be any way to do this via the mailbox.
This has greatly optimised the 12 bit capture code and the alpha channel can be used to vary the output intensity of the RGB values without altering them which is useful for variable scanlines and OSD dimming.

An overlay bit on GPIO 24 has been added in 12bpp mode to support the flash floppy OSD output.
When FF video is connected to GPIO  24, the video signal is overlaid on the image similar to the normal OSD.
This option only works with c0pperdragon's board at the moment.
It could be extended to 12bpp with the CPLD board but not 3 or 6 bit mode so I'm looking at a serial comms based option instead.

Mode 7 calibration has been improved so that an error mode which resulted in a clean image but with character verticals one pixel wider than normal can be detected. When this failure mode is detected, 1000 fake errors are added to the diff tables for each of the 6 offsets so for an otherwise clean image you will end up with 60,000 errors (1000 x 6 x 10 frames)

I think it's time for another dev release, potentially as a release candidate.
The only changes I intend to do before the next stable release will be bug fixes and profile updates


hoglet67:
> I think it's time for another dev release, potentially as a release candidate.
> The only changes I intend to do before the next stable release will be bug fixes and profile updates

Here you go:
https://github.com/hoglet67/RGBtoHDMI/releases/tag/20201126_2d102ae

PR #197, opened 2021-03-27
Amiga Video Slot V2 layout

LinuxJedi:
The main differences from the V1 slot adaptor are:
- ICs moved to more appropriate locations
- Fixed floating inputs on U5
- Added ground pour and more grounding points for RGB and Pi
- Manually rerouted everything (fewer signal vias)
- Added option to use U1 for the inverted C1, C3 XNOR (which makes it an XOR)
- Added RC based delay option
- Input capacitor added to the regulator
- Rounded the corners a little more

I've also added build notes in the README.md in the V2 directory.


kicad_AmigaAdapter/VideoSlot/V2/README.md
```diff
@@ -0,0 +1,42 @@
+# Amiga Video Slot Adapter V2
+
+This adapter builds on the ideas from the Amiga Denise adapter V2, adapted for use in big box Amiga video slots. Version 2 has been modified for reliability and to provide timing options.
```
Bloodmosher, line 3: Why not explain that this is derived from the video slot adapter V1?

kicad_AmigaAdapter/VideoSlot/V2/gerber/videoslotadapter-F_SilkS.gbr
```diff
@@ -0,0 +1,2183 @@
+G04 #@! TF.GenerationSoftware,KiCad,Pcbnew,5.1.9-73d0e3b20d~88~ubuntu20.10.1*
```
Bloodmosher, line 1: If this silk screen includes "Modified by LinuxJedi" I would suggest sticking with the precedent that c0pperdragon set and I also followed for the V1 of the video slot adapter, of leaving credits to a single person off the silk screen.

LinuxJedi: Fair enough, I'll fix that up shortly.

kicad_AmigaAdapter/VideoSlot/V2/README.md
```diff
@@ -0,0 +1,42 @@
+# Amiga Video Slot Adapter V2
```
Bloodmosher, line 1: Some additional thoughts that I think are worth covering before completing this as this PR will establish precedent.

1. There are a LOT of other boards out there now, some even more radical than this one. Many of those will just remain as forks of the project and never merge into main. That should be totally fine. So it is a good question of what should be here versus a fork.

2. What is version "N +1" is also a good question, versus just making another folder for an "alternate" build. The answer cannot be "whoever made the last radical redesign is the latest".

3. I like the idea that main contains a build that is easy to do, has only one hardware setting (which Denise), and can provide great picture quality for any Amiga it is put in. Right now the V1 is that if you add the 47pf capacitor (doesn't hurt and most likely helps). You might need to also tweak software, but not in all cases. So with that single change, we have an easy to build and configure board that should work for most if not all Amigas.

4. This board has a LOT of changes to it. Some seem arbitrary, whereas other seem warranted. It is hard to know what impact any single change has, for example the fill zones or the chip re-arrangement. What do we do with the next PR to come along that re-arranges them "just because"?

5. On the fill zones, I personally think this just makes future expansion more difficult, e.g. adding either more Pi hardware to the board, or making use of the other video slot signals (e.g. PPD, PBUSY, PPOUT, PSEL, etc.). There are some interesting possibilities there that if I were starting on, the V1 board is just simpler to update. Given what can be achieved with the V1 board in terms of picture quality, they don't seem required.

6. As of the moment this board has very little testing by others. I've got one on order and will be happy to share my experience. But is it ready to go in and look like it is a replacement for V1? I'm not sure yet, especially given the above and thoughts on future expansion. Perhaps at this time either a fork or an "alternate" sub-folder is better?

LinuxJedi: Given your above concerns I'll remove this PR and keep a fork instead.

LinuxJedi:
Closing due to concerns from original slot designer.

kicad_AmigaAdapter/VideoSlot/V2/README.md
LinuxJedi: To answer your questions though.. 1, 2 and 4 would typically be the job of project maintainers to sort out and give guidance on how such things should work.  3 fair enough, that is why I added a space for the 47pF. 5 KiCad deals with that automatically. 6 it has gone through extensive testing, but how much testing qualifies enough is another question entirely.

hoglet67:
I haven't been actively following the Amiga work, so it's hard for me to know what to merge. This pull request is a great example of this. I will generally wait several days before merging stuff, to allow errors to be discovered, or folk to raise objections. But actually it feels wrong for me (or IanB) to be the arbiter of this stuff.

My own view is that it's probably better for each new hardware design to go into it's own git repository. Then it's down to the 
board originator to decide what changes should be merged. They can also control how branching is used.

The other consideration is that I'd like to keep the main RGBtoHDMI repository as small as possible. This pull request was 113,497 lines. Stuff is git is additive, so if you are not careful, over time the repository becomes enormous.

(Part of this is people not understanding the kicad file format. The fp-info-cache file, for example, does not generally need to 
be committed. That file alone is 82,000 lines in this case.)

So my suggestion would be to start using seperate repositories for new hardware projects.

Dave (hoglet)

LinuxJedi:
> My own view is that it's probably better for each new hardware design to go into it's own git repository. Then it's down to the
> board originator to decide what changes should be merged. They can also control how branching is used.

That is fair enough, I couldn't immediately see any information indicating this. But I'll take it under advisement for future.

> (Part of this is people not understanding the kicad file format. The fp-info-cache file, for example, does not generally need to
> be committed. That file alone is 82,000 lines in this case.)

That is my bad, I usually use .gitignore to manage that so that adding a dir doesn't add everything.

> So my suggestion would be to start using seperate repositories for new hardware projects.

Will do, many thanks for your input.

PR #199, opened 2021-04-14
V1.01 layout updates to include 47pf capacitor, ground extra floating pins on U5

Bloodmosher:
V1.01 Added spot for 47pf capacitor (C6) to ensure no sparkling on image, and grounded unused inputs on U5 per data sheet recommendation.


Bloodmosher:
@hoglet67 friendly ping on this one...

PR #201, opened 2021-04-20
Updates to support single button mode and FFOSD with CPLD plus new recovery mode

IanSB:
Fix single button mode issues with recovery mode when using CPLD
Add FF OSD support in CPLD 12bpp mode only (FFOSD video input on old MUX pin - GPIO 24)
Slightly improve RPI4 support (still needs work)
Add new recovery mode for 8 and 12 bit boards

I realised that as 8 and 12 bit boards can be distinguished from the earlier 3 and 6 bit boards due to the identity resistor R7 it would be possible to add a separate CPLD recovery mode for those boards which offered all 3 of the 8/12 bit CPLDs (BBC/RGB/YUV) during initial setup with a blank CPLD so if you needed RGB or YUV you wouldn't have to do a double program.
The older 3 and 6 bit boards should still get the original recovery menu with the two BBC CPLD options.
I'm not sure which recovery mode the Atom board will display so it still supports Atom reprogramming via debug in the new recovery menu.


hoglet67:
Ian,

I probably won't be able to look at this until the weekend as I'm in the middle of refurbishing my work room (fitting new desks etc), so everything is a bit of a mess.

Dave

hoglet67:
Merged into both dev and master

hoglet67:
(Tested on the Atom, no issues so far)

PR #218, opened 2021-07-02
Add TI-99/4A profile

hanshuebner:
This profile works with a European (TMS9929A) TI-99/4A that has a
tms-rgb board (https://tms-rgb.com/) fitted.


hoglet67:
Just a quick note to say we are still considering TI-99/4A support.

Regarding this pull request, IanB said the following:

The TI-99/4A for the TMS-RGB will not work properly as the analog RGB thresholds aren't set up. It will likely produce an approximate result but not accurate colours. I now have one of the TMS-RGB boards so I will look into this further when I can find the time to fit it to my TI-99/4A.

PR #222, opened 2021-07-16
Amiga Video Slot V1.1

Bloodmosher:
The updated V1.1 board should fix sparkling pixel noise on even the most demanding video patterns (see https://github.com/Bloodmosher/Amiga-VideoSlot-RGBtoHDMI for an example). Thanks to LinuxJedi for demonstrating that alternate layouts have an impact on this. While it is difficult to pinpoint exactly which combination of changes is absolutely necessary, of the numerous experiments I have tried with variations on path length, eliminating through-holes, copper fills, regulator placement, etc. the predominant issue seems to be placement of U1 relative to the other chips. With this update the "extra" 47pf capacitor should no longer be necessary and the image should be "sparkle free".


Bloodmosher:
@hoglet67 Friendly ping - this should  be a good candidate for a 'final board' for this repo (at list for this design based on c0pperdragon's original Amiga board); it's undergone a lot of testing to ensure no sparkles on the image after numerous iterations. Also please see my comment on PR https://github.com/hoglet67/RGBtoHDMI/pull/224 - I don't think we should put untested boards in this repo and should stick to your previous guidance on other board variations residing elsewhere.

PR #224, opened 2021-07-18
Amiga 2000 Video Slot Adapter, which mounts the Pi to the Slot Bracket

jan-hofmeier:
This is Based on the V2 Version of the Denise Adapter. I adapted it to the Video Slot of the A2000, similar to what
Bloodmosher did with his Video Slot Version. The Main difference is, that this puts the HDMI port of the Raspberry right against the Back of the Amiga, so no internal HDMI extension is required.

I didn't find time in the last 6 months to test it and I probably don't have time to test it the next few months. I also noted in the README that it is still untested. But as its based on a working design this should work too. I am sorry for that. I am not sure if this is a requirement for being accepted.


Bloodmosher:
Please let's not put any untested boards in this repository- just leave in your own fork or a separate repo. A lot of effort has gone into testing the most recent update (https://github.com/hoglet67/RGBtoHDMI/pull/222) to ensure there are no sparkles in the image, and any untested  board is very likely to have problems with this. In general there are a dozen or so variations on this board now and hoglet67's guidance has been to put them in separate repos.

hoglet67:
Jan,

Thanks for this pull request.

Managing a central repository of boards seems like an unnecessary overheard.

I would generally prefer third-party boards to live in their own repository.

I'm happy to link to this from the Wiki as and when you get around to testing it.

Dave

PR #229, opened 2021-08-04
Update RGB CPLD to V94, Improved calibration, switchable timing sets, Apple IIGS, Apple Lisa support

IanSB:
The RGB CPLD has been updated to V9.4 which has the following changes:
Test for Apple IIGS Super Hires Modes (Monitors a signal on the VSYNC input which toggles 1/0 in SHR mode (16.3Mhz pixel clock) and is '0' in other modes (14.3Mhz pixel clock)
Add switchable leading/trailing sync edge same as the YUV CPLD
Add switchable delay to leading edge (required for Apple Lisa)
Add 9bpp capture mode with blanking on B3 (Required for Atari ST which blanks the signal in the analog domain after the DACs)
Add 1bpp capture on the VSYNC input (Required for Atari STE so that both 12bpp lores and mono hires can be picked up)

Automatic filtering out of flashing cursors during calibration.
Previously only BBC/Electron cursors were filtered out by hard coded line numbers but now flashing cursors on any system are automatically detected and filtered out during calibration so you can get zero errors when calibrating.
The software looks for a single run of consecutive lines with errors up to a maximum of 12 lines (6847 on CoCo) when the rest of the image is error free and if detected those lines are ignored. As a special case, in the BBC and Electron profiles that limit is reduced to 2 lines (which is all that is required) so the filtering has less impact in mode 7.
You could simulate the Tandy CoCo flashing cursor on an Atom to test this by flashing a full height block.

Improve selection of sampling phase with auto calibration:
Previously selection was done by searching for the lowest error value, preferably zero with zero either side. This worked OK but using higher multiplier settings where there were many 0 error sampling phase values, the actual position selected relative to the worst case error result was somewhat unpredictable.
Now, where possible the sampling phase is selected 180 degrees from the worst case value: 
i.e. half the multiplier is added to the worst case sampling phase value and that position is tested for 0 errors with 0 either side and if that is OK that value is chosen, if not then it reverts to the old method.
There is also a new setting in the sampling menu:
Calibration Range: with options (Auto or 90/270 degrees)
In Auto mode it selects 180 degrees as described above.
In 90/270 degrees mode it selects a value 90 or 270 degrees from the worst case value (i.e. adds or subtracts a quarter of the multiplier). This option is required for the Atari ST because the 320 pixel and 640 pixel modes are 180 degrees out of phase. i.e. the best sampling phase for 320 is the worst for 640 and vice versa. This setting allows a compromise value that works for both.
Note the "Calibration Range: " setting has a different meaning in mode 7 where it is used to control auto selection of a higher multplier as previously.

The selection between the two sets of geometry/sampling values for modes0-6 and mode7 has been generalised and they are now referred to as "Timing Sets":
This was required for Apple IIGS support which like the BBC has two different pixel clocks
There are now several ways to switch between the sampling sets using the "Auto Switch" setting:
Off - no sub-profile or timing set switching
Sub-profile Only - Switch to different sub-profiles in a profile folder
Sub + BBC Mode7 - as Sub-profile plus select Timing set 2 if mode 7
Sub + Vsync - as Sub-profile plus select Timing set 2 if vsync is high (not currently used by any profile)
Sub + AppleIIGS - as Sub-profile plus select Timing set 2 if vsync is toggling high/low (always low = Set1)
Sub + IIGS + Manual - as Sub + AppleIIGS plus manual override on SW3 (toggles)
Sub + Manual - as Sub-profile plus manual toggling between set 1 & 2 with SW3 (Used by Acorn Communicator profile)

There is now a prompt to save configuration after running the auto config and a reminder that you have to use calibration to fix noise  is printed as part of the startup message. (A lot of people don't seem to realise that).

D1 & D2 LED status when no sync:
D1:
Off = no sync
flashing = attempting to genlock
On = genlocked
D2:
Off = Timing Set 1
On = Timing Set 2 or No Sync (ensures at least 1 led is lit when no sync so a powered board nearly always has one led lit)

When in NTSC artifact mode SW3 cycles between the four artifact quadrants so you can get the colours correct. Primarily for the Tandy CoCo machines which randomly vary the phase by 180 degrees on power up.

Dragon/CoCo 6847 palettes have been updated to be more accurate.

Various profile updates and bug fixes.

As there have been significant changes to the calibration and mode timing set switching, this needs significant testing to ensure no new problems have been introduced so can you look at that.
Hold off on a new release for now until some more testing has been done.


hoglet67:
Ian, I'll spend some time on this tomorrow, once our visitors have departed.

PR #235, opened 2021-08-31
Added a profile for the Swedish computer ABC80

sweproj:
I have added a profile for the Swedish computer ABC80


hoglet67:
Just a quick note to say we are still considering ABC80 support.

Regarding this pull request, IanB said the following:

The ABC80 also looks problematic. The pixel clock is set to 12Mhz but the schematic indicates 11.9808Mhz:
https://www.abc80.net/old/ver1/luxor/abc80schema/abc80schema_3.gif
Also the palette is set to the TEA1002 which is not really the correct palette for a mono machine
It will probably work due to the clock tolerance setting but should be corrected for reliable operation.
(There is also an ABC800 with 12Mhz pixel clock and RGB)

sweproj:
Hi

Thanks for the comment. I will update the clk. and make a new pull request but before this, just a comment/question. I have tried the monochrome profile as well but the text gets quite dim even with, i will try to investigate more before the new pull
BTW this was my first pull request ever so sorry if its not correct
Per

IanSB:
Thanks. This pull request has been integrated manually.

PR #242, opened 2021-10-22
Added NEC PC-98 and Epson clones profiles.

cuba200611:
These profiles have been tested with an Epson PC-486GF, which is a PC-98 compatible machine.


IanSB:
Thanks. This pull request has been integrated manually.

PR #251, opened 2021-11-17
Full Pi zero 2W support with capture in the GPU on VPU core 2

IanSB:
Full Pi zero 2W support with capture in the GPU. The source can optionally be compiled to run with the old ARM based capture but that is only really useful when running on the original Pi zero or Pi 1.

The VPU code (videocore.s) is in the src folder with the other source files and can be built by running buildvc.sh in the scripts folder.

There is now an additional script 'configure_rpiA.sh' which creates 'kernelrpiA.img' which is an ARM capture version of the pi zero / Pi 1 code and running release.sh will also create that. Currently it is only used by the Pi1 in config.txt because the core clock on that is 250Mhz and cannot be increased beyond 300Mhz and you get a better result with the Arm only code.

You can optionally run the original Pi zero with the ARM build by editing config.txt or renaming the file.

It is possible to build ARM only versions of the Pi2 and Pi3/zero2W for experimentation by uncommenting USE_ARM_CAPTURE in defs.h but they are very limited with no 9/12bpp support, no double height capture (the setting is ignored as it causes glitches) and mode7 is also slightly glitchy so I didn't include them in the release set

I've decided to remove very old 3 bit CPLD support (V6 and earlier) from the GPU capture build but it should still work in the ARM build so anyone using an old non-reprogrammable CPLD will have to edit config.txt or rename.
(Actually old CPLD support is completely broken at the moment and must have been for some time but I only just noticed. It is likely an issue with the register bit order in cpld_rgb.c that happened when I last made changes to the CPLD)

I have released this as Beta43

Can you check that the Pi2 works with this.


hoglet67:
> Can you check that the Pi2 works with this.

The old Pi 2 seems to be working fine. Initially I was getting No SYNC, but then I realized that I was using a older "Six Bit Issue 2" board with the BBC 7.3 Firmware. Upgrading to the latest 7.9 firmware resolved that. I musg go back and check the original 3-bit boards are still working.

I also tried the Pi 4, and while that works OK in MODEs 0-6, MODE 7 is very jittery (~half a character of jitter through out the screen).

IanSB:
>I also tried the Pi 4, and while that works OK in MODEs 0-6, MODE 7 is very jittery (~half a character of jitter through out the screen).

Yes, that's what I meant. If you reduce the max width in the geometry menu a few clicks that stops so I think sometimes it's taking too long to capture and missing the start of sync due to latency.

IanSB:
>Initially I was getting No SYNC, but then I realized that I was using a older "Six Bit Issue 2" board with the BBC 7.3 Firmware. 

I think something is broken with old CPLD support as mentioned.
Not sure when that happened.

IanSB:
BTW what Framebuffer address are you getting on your hardware as there seems to be two areas in use depending on Pi version and maybe memory size.
I get:
DE000000 on Pi zero and zero2w
FE000000 on Pi 3B and Pi4
I've marked the upper part of both areas as cacheable for mode 7 decoding but not sure if that's the only areas in use.

IanSB:
Your memory map shows `# 0x3F000000 1008MB - Start of GPU Memory` so is that on your Pi 4?
```
...
# 0x0E800000 232MB - Top of cached RAM
# 0x3F000000 1008MB - Start of GPU Memory
# 0x40000000 1024MB - Top of 1GB of physical RAM
```
PERIPHERAL_BASE  is   0x3F000000 on the Pi2 & Pi3

hoglet67:
That was me hypothesising what the memory map might look like on machine with 1GB of RAM and gpu_mem set to 16.

On a Pi 4, MODE 1024,1024,32 (8MB) ends up with the firmware returning a FB Addr of 0xFEBFB000->0xFF3FB00.

We currently ignore the top 2 bits so use 0x3EBFB000 and this seems to work.

I'm still confused about the 32-bit memory map on the Pi 4. Maybe I need to read the BCM2711 data sheet again.

PR #256, opened 2021-11-30
Some minor bug fixes and improvements plus universal binary support

IanSB:
Slightly improved mode7 calibration
Add Swedish computer ABC80, NEC PC-98 and Epson clones profiles
Fix garbage on right hand side of screen in teletext mode when using GPU capture
Merge support for Pi zero, zero2W, 1, 2 & 3 into one binary
Pi4 support now mostly working, only genlocking and screencaps not working yet.
Support old CPLDs in GPU capture mode
Updated build scripts

Note the builds are now:
kernelrpi.img - GPU capture for Pi zero, zero2W, 1, 2 & 3
kernelrpi4.img - GPU capture for Pi 4
kernelrpiA.img - ARM capture for Pi zero, zero2W, 1, 2 & 3
kernelrpiA4.img - ARM capture for Pi 4
(Plus their debug variants)

There is also a "ConfigA.txt" and renaming that to Config.txt will run all the ARM versions
Note the ARM versions are mainly there for reference and comparison and may be removed eventually
The ARM build should run as well as the original ARM build on the Pi zero although you will probably have to increase the overclock back to previous levels in 12bpp modes.
The ARM build doesn't work at all well on the zero2, Pi 2 and Pi 3. In order to get usable images, double height is disabled and teletext deinterlacing is set to weave when running it on those Pi models. 12bpp modes don't work at all.

The Pi 1 and Pi 2 are both overclocked so that 12bpp capture works as well as other models but there is a setting in config.txt (auto_overclock) that controls this. 
The Pi 1 is overclocked from cpu=700,core=250 to cpu=900,core=400
The Pi 2 is overclocked from core=250 to core=400.
Not sure if it is better to have this enabled or disabled by default.

I have released this as beta44


hoglet67:
Ian, can I just check what version of ARM GCC Embedded you are using to build the beta builds?

I'd like to make sure I'm using the save version.

IanSB:
arm-none-eabi-gcc-7.3.1 (GNU Tools for Arm Embedded Processors 7-2018-q3-update) 7.3.1 20180622 (release) [ARM/embedded-7-branch revision 261907]

Issue #253, opened 2021-11-19
Pi4 remaining issues


IanSB:
I though I'd document the remaining issues with Pi 4 support to help in tracking down the required info:

1. Genlocking doesn't work.
This is due to the HDMI PLL registers moving or changing
Also the interrupt controller has changed so the Vsync flag can't be read (used for genlock and displaying the red sync bar)

2. Screecaps don't work
lodepng genrates error 55

3. 16bpp display modes don't work properly because the mailbox will only select 5:6:5 RGB and the display list gets modified to switch this to 4:4:4:4 ARGB. The display list and associated registers have moved or changed so that trick no longer works.
This could be worked around by reverting to the original 5:6:5 capture code. The reason for the change was to speed up the capture loops as almost no logical masking and shifting was required for 4:4:4:4 ARGB compared to 5:6:5 RGB.
However due to the new GPU capture code the loss of speed by reverting may not be an issue especially if it is only done for the Pi 4
 
4. The PLLs need further experimentation. 
I've managed to get it working using PLLD but all the PLLs seem to be in use so this needs to be looked at further as it currently runs the PLL higher than it's default rather than lower (running lower causes the Pi to lose display or lockup.)
 
5. Mode7 is glitchy
Mode 7 is glitchy even with GPU capture. However this can be fixed by reducing the capture width by a few pixels (it's already wider than the active screen so nothing is lost).
This is very likely due to higher memory latency than on the other Pi versions.

IanSB:
Update:

Issue 1: The interrupt register for the SMI interrupt (fake vsync) is now in IRQ0_PENDING1 on the Pi4 at offset 0x00B204. After changing that the vsync red bar now works (genlocking still not working)

Issue 5: Mode7 is now OK it was using the cached screen option but the memory area wasn't actually set as cached which caused the glitches.

IanSB:
Issue2: Screencaps are now working after changing the compiler options to target arm v6 (like the universal Pi0-3 build) instead of arm v8 so the compiler output for Arm v8 has some issues, perhaps something isn't set up correctly like unaligned access.

IanSB:
Update:

16bpp display modes now work properly:

The display list has moved from offset 0x402000 to 0x404000
Also the structure of the list has changed so that the start of video entry is at offset 6 instead of offset 5
Finally the values representing the bit order in the first word of the list have changed:
#define PIXEL_ORDER 2   // ABGR in BCM2711
[...]

PR #257, opened 2021-12-04
Genlock working on Pi 4 + minor bug fixes

IanSB:
Genlock now working on Pi4
Minor bug fix for refresh rate display
Improved use of PLLD which doesn't slow down the SD card as much
Fix compiler 10.3 string warnings


IanSB:
See #253 for an update on the corruption issue

PR #299, opened 2022-09-04
Add Amiga 3000 PAL profile

grovdata:
Tested with [BloodMosher](https://github.com/Bloodmosher/Amiga-VideoSlot-RGBtoHDMI) board, works good in PAL A3000.


IanSB:
I have added this to the most recent release although it has been manually merged due to some changes in folder structures.

PR #310, opened 2023-02-19
Add support for line doubled capture with improved integer scaling

IanSB:
These changes allow capture of line doubled sources with half integer scaling values e.g. 2.50

To use
Set the vertical min and max height to half of the required value e.g. for 480 lines set them to 240
Set FB Size to Double Height or Double Height+Width depending on how your width is setup
Set Video Type to "Line Doubled" (this setting is ignored if Double Height is not enabled)

After this the scaling is calculated using the above settings but the capture actually works as if min & max height are double the displayed value (e.g. 480 for a 240 displayed value) with FB Size Double Height disabled

Note this has no effect on horizontal. e.g. To get similar results for 640, use half the pixel clock rate and 320 pixels with Double Width enabled.

### DIFF
--- a/src/defs.h
+++ b/src/defs.h
@@ -558,10 +558,11 @@ typedef struct {
 #define  VSYNC_POLARITY                6
 #define  NUM_VSYNC                     7
 
-#define  VIDEO_PROGRESSIVE 0
-#define  VIDEO_INTERLACED  1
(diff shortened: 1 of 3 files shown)


hoglet67:
Cheers Ian, that works very nicely.

Dave

PR #311, opened 2023-03-01
New profile selection menu plus higher resolution monitor support

IanSB:
The main change is a new "Select Profile" menu option:

When selected, a sub-menu of manufacturers is displayed and when one of those is selected, a further sub-menu of profiles is displayed for all CPLD variants. The profiles for the current CPLD are displayed normally and the profiles for other CPLDs are displayed with an appropriate prefix (BBC, RGB, YUV). 

If you select a profile for the current CPLD it returns to the main menu with that selected.
If you select a profile for another CPLD it sets that as the boot up choice for that CPLD and then reprograms the CPLD followed by a reboot.

Other changes:
Improved support for greater than 1080p resolution monitors.
2560x1440 is now detected when in auto detect mode although it required a workaround as that resolution isn't handled in the Pi firmware
When a 4K monitor is connected it selects 1920x1080p. This results in some interpolation by a 4K monitor even in integer mode. Previously it selected 3840x2160@30Hz which looked sharp but dropped every other frame and wouldn't genlock. 

I have added a new manual resolution of 1920x2160 which is intended as a compromise for 4K monitors which results in interpolation in the horizontal direction only. However as it is a non-standard mode it is not chosen by default.

Interestingly this 1920x2160 resolution works on my 2560x1440 monitor as well and looks great as the monitor is dropping lines to scale the vertical down rather than interpolate, filling the screen and it happens that it is dropping the same number of lcd pixels per source pixel.
The only time it looks bad is if scanlines are enabled because the pixels being dropped are at different points in each source pixel and that makes the output look a mess.

(Pi models except the Pi4 can't support 4K @ 60Hz so I may add a true 4K option for the Pi4 as well)

Also on the Pi0 & Pi1 (not Pi02W) when in 12bpp modes the frame buffer is cropped to the exact capture size with the overscan being adjusted to compensate. This is due to memory bandwidth problems resulting in glitches with certain combinations of profile, scaling and screen resolution.

Finally, I refactored the code that handles the parameters for the feature menus and that eliminated a lot of unstructured variables and their associated get / set code and also makes adding feature menu options simpler. 

I still have further testing to do but can you give this a try before I do a release build.
Some things to consider:

Should 3840x2160@30Hz or 25Hz be supported?
This will require genlock changes but the output will look sharp at the expense of half frame drops.

For the time being I have moved the Test50Hz option to the Info menu to keep the main menu the same length but it might make sense to put the "Update CPLD menu" in there as well as most cpld reprogramming will now be done via the select profile menu.

Any thoughts on the best option for this?
i.e. both of the above in the info menu or one of them back in the main menu?
(I originally put the test 50Hz in the main menu to make it high profile to encourage users to try it)


IanSB:
I've also added a "Recently used" option which keeps track of the 10 most recent profiles that have been selected from the select profile menu for quick access.
Not sure if 10 is the optimum number.

IanSB:
>I'm a bit tied up with some other stuff today and tomorrow, but I'll try merge this and do some testing over the weekend.

Thanks

I've been fixing bugs and made a few improvements for 4K support so genlock now works in 25-30Hz modes although a warning is displayed if you select one of those resolutions.
4K options are now:
Auto: 1920x1080p
Half 4K(1920x2160) - non standard and may not work on some monitors
Full 4K(25-30Hz)      - full 3840x2160 resolution but at half frame rate

IanSB:
Could you also test this with your superboard?
I made some changes to the profile and also added a new option "YUV Pixel double" specifically for that. When enabled it drops alternate pixels and doubles up the remaining ones which makes the font solid. 
(I think you tried something similar with half rate sampling but that would often auto calibrate to black as it locked onto the wrong part of the font)
I also made some changes to the voltage levels and I'm not sure if they are optimal as I had to replace the video pot on mine so I don't have a factory setting.

Normal:
capture1

YUV Pixel double enabled:
capture2

hoglet67:
> Could you also test this with your superboard?

I'm running the new software now.

On the Superboard the profile worked, but the vertical screen position was too high. Reducing V Offset from 18 to 2 resolved that.

I'll update my other two devices, and ley you know if I hit any issues.

IanSB:
>On the Superboard the profile worked, but the vertical screen position was too high. Reducing V Offset from 18 to 2 resolved that.

I tweaked it to more centre the video when used with CEGMON which only writes to 25 of the 32 lines and the bottom one of that 25 is one line lower. The standard SYN600 monitor writes to 28 lines but the top 3 or 4 of those are usually not viewable in the vertical sync and blanking period (the text just scrolls up through them). 
The Superboard generates video for the entire video frame including H sync and  H blanking periods to simplify the video circuitry and relies on those mostly being left as space chars otherwise you get clamping issues and flyback lines.

With the new profile selection system I suppose I could add separate profiles for the different versions now that it is no longer desirable to keep the number of profiles to a minimum to make selection faster.

>I'll update my other two devices, and ley you know if I hit any issues.

Have you had a chance to test the others?

BTW I have just added a soft reset feature for cases when the board is plugged into a Pi without any reset mods.
Pressing SW1 and SW3 simultaneously will cause a reboot.
The hardware reset is useful when developing in case of crashes etc but sometimes resetting is useful to clear out any tweaks, update the SD card or recover from a screen mode that doesn't work using menu-reset and in those cases it usually doesn't require a hardware reset.

If two diodes are added between the reset button and SW1/SW3 (banded end towards the reset button), the reset button appears to work normally even when not connected to the reset line.

The only major difference is that to enter CPLD recovery mode you first have to press & release reset (or SW1/3) before repressing all three switches. With the hardware reset you could hold down the three switches before pressing reset.
Menu-Reset still effectively works in the normal way: i.e. hold down Menu (SW1) then press reset (or SW3)

I think I will do an updated CPLD board with these mods as it will become more desirable when used with the Pi zero 2W due to the messy nature of the hardware reset mod.

hoglet67:
I'll commit it tomorrow then.

PR #324, opened 2023-05-16
Improved help messages and profile updates

IanSB:
The changes in this pull are minor and mainly some improvements to the help messages and some profile updates including one for the Exidy sorcerer

I think it's time for a new stable release.

You have already merged most changes into dev since the last stable release so I will summarise all significant changes since then:

This release has a major new feature, a "Select Profile" menu option that makes changing profile a lot easier.
After selecting the option you are presented with a list of manufacturers and after choosing one you are then presented with all profiles for that manufacturer, including profiles for the other CPLD versions as well as the current CPLD.
Profiles that require a different CPLD firmware are prefixed by (BBC), (RGB) or (YUV) as appropriate.
Selecting a profile with no prefix will immediately switch to that profile
Selecting a profile with a prefix will reprogram the CPLD, reboot and select that profile.
There is also a recently used option that keeps a record of the last ten profiles selected.

Other changes:

Add support for c0pperdragon's new Commodore 64 VIC II plug in adapter which outputs a special 'lumacode' signal on the modulator Phono output that RGBtoHDMI can decode into bit perfect HDMI.

Auto detection for 2560x1440 monitors, (previously the Pi would not auto detect this resolution)

Some support for 4K monitors:
The pi zeros don't support 4K @50/60Hz so by default RGBtoHDMI will select 1080p@50/60Hz as appropriate when connected to a 4K monitor. This is not ideal as the monitor will interpolate that to 4K resulting in soft edges.
However there are two manual modes to try:

Full_4K@25/30Hz
This will give 4K resolution but at half frame rate (which the zero supports) although every other frame will be dropped giving a 'film effect' on smooth motion.

Half_4K 1920x2160@50/60Hz
This is a compromise resolution which is the same height as 4K but half the width and again the Zero can generate that.
This results in integer scaling vertically and interpolated scaling horizontally but as it is a very non standard mode, a lot of monitors and TVs may not support it and give a blank screen etc.
Also it may be displayed as a narrow image in which case you need to change the monitor's scaling to fill the screen (not RGBtoHDMI scaling).

Soft Reset
Pressing SW1 and SW3 together will cause a reboot and this can be useful if the board is plugged into a Pi that doesn't have any reset mods made to it.

No CPLD detection
If you boot up the software with no CPLD board plugged in you will get an error message and the main menu displayed and this can be useful to confirm the software is installed and booting properly if you have a problem when the CPLD is plugged in.

Help Menus
There are several help pages in the info menu describing button functions, calibration, NTSC artifacts and common problems plus where to download updates.

Supported Profile List
There is a menu option in the info menu to save supported profile list for current hardware as "Profile_List.txt" for easy reference. (This list will be different depending on analog or digital interface)

Add warning about pre-programmed CPLDs to the CPLD recovery menu. (Using CPLDs with other firmware already installed can cause reprogramming problems and this describes the workaround)

Add 8 level mono palette

Add support for line doubled capture

Sync detection moved to GPU (Improves performance at high pixel clock rates)

Fix issues stopping very high resolution sources from being captured (864x1024@60Hz 3BPP has been tested)
Fix sync polarity detection when auto sub-profile switching

Add Commodore 64 lumacode profile (see above)
Add Exidy Sorcerer profile
Add NEC PC 16-01 profile
Add Meldas CNC9 profile
Add Amiga 3000 profile
Add EGA Wonder 800+ profile
Add Atari ST no blanking profile (for earlier adapter board)

Update Spectrum Harlequin and Nuvo profiles
Add new test profiles
Plus many existing profiles updated


hoglet67:
Merged back into master, and a new stable release created, using ARM GCC 12.2:
https://github.com/hoglet67/RGBtoHDMI/releases

PR #485, opened 2026-07-26
Add Commodore A2024 framebuffer decoding and autodetection

reinauer:
Add native support for the Commodore A2024 monitor protocol.

The Amiga does not generate the final high-resolution raster directly. Instead, it sends ordinary RGBI fields containing sections of a larger four-level grayscale framebuffer:

  - Four fields for the 15 Hz transfer format
  - Six fields for the 10 Hz transfer format
  - Approximately 1008x1024 output for PAL
  - Approximately 1008x800 output for NTSC

Each RGBI sample carries two adjacent 2-bit grayscale pixels. In-band Hedley control words identify the transfer format, destination panel, pixel depth, and effective width.

## Implementation

  - Capture A2024 command and data lines in their native 12-bit layout.
  - Validate the complete control signature to avoid false detection on ordinary Amiga video.
  - Decode four-field and six-field transfers into one persistent framebuffer.
  - Preserve logical grayscale palette indices without applying the generic 9/12-bit palette permutation.
  - Keep A2024 capture in a late-linked assembly object so the normal PAL capture hot path retains its established layout.
  - Require two valid command fields before entering A2024 mode.
  - Require four genuine misses before returning to the previous normal Amiga subprofile.
  - Preserve valid fields that require a deferred depth-table update.
  - Save transition counters and raw command samples in `A2024_Probe.bin` when diagnostic logs are requested.

A2024-aware autoswitching is enabled for the generic Amiga, A2000, and A3000 profile families. Separate profiles remain necessary because the detector selects an existing subprofile to obtain the correct palette, geometry, timing, sampling, sync, and frontend-specific settings.

The `50Hz` and `60Hz` profile names describe the PAL or NTSC source timing. Separate 10 Hz and 15 Hz profile files are unnecessary because that transfer format is selected from the in-band control words.

## Build compatibility

The series also updates the build for current tools:

  - CMake 4 policy and cross-compilation support
  - Explicit C and ASM compiler selection
  - Static-library cross-compiler checks
  - Unix Makefiles instead of the deprecated CodeBlocks generator
  - Exact function prototypes required by GCC 15's C23 default


reinauer:
image

